PR #26, opened 2015-03-25
Add *Actual speed* feature request

ViliusKraujutis:
When playing videos in really high speed (eg. 100x, yes, with some #solareclipse videos I've done that), it's actually played slower. Maybe my HW is not capable to render video stream that fast or what... I wish to know what was actual playback speed, because it looks it was only ~10x faster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ Note that you can customize these shortcuts in the extension settings page. Also
 ### Wishlist
 
 * Make it not suck with Vimeo content: "s" shortcut and button clicks
+* Show actual playback speed
 * Your awesome feature here...
 
 


igrigorik:
100x!? Wow! So, I'm not sure we can do much here in the extension... 

As in, there is no real way of figuring out the `x`, short of decoding and monitoring each frame, or some such.. which would be incredibly heavy and costly. I'm guessing different platforms will have different limitations about what max speed they'll play -- e.g. I know Chrome cuts off audio after certain threshold. 

/cc @samdutton

jordn:
Chrome cuts off audio above 4.0x and below 0.5x. Just doing some crude stopwatch timings on Youtube there seems to be an upper limit of playback speed at ~16x. Going beyond that it has no effect on the actual playback speed. This is Chrome 41 on a 2013 Macbook Air but I believe my experience on other devices has been the same.

@igrigorik Is there a way to confirm from the Chrome team what the upper limit is? We could conceivably set an upper limit on this extension to limit expectations for the user.

queuebit:
@jordn I would probably start here: <https://code.google.com/p/chromium/issues/entry>.

igrigorik:
Video playback speed is [0.0625, 16]: https://code.google.com/p/chromium/codesearch#chromium/src/media/blink/webmediaplayer_impl.cc&l=64

Audio is muted for rates outside of [0.05, 4]: https://code.google.com/p/chromium/codesearch#chromium/src/media/filters/audio_renderer_algorithm.cc&l=49

My proposal is to to add guards in videospeed for [0.05,16].. outside of that range neither video nor sound works.

jordn:
Neat – thanks for linking to the source. Proposal sounds sensible.

Issue #34, opened 2015-04-03
Video speed retained ignoring that rememberSpeed setting is false


jordn:
The speed persists between videos, even though I am expecting the default behaviour of not remembering the speed.

As far as I can tell the extension no longer pays attention to this setting.

jordn:
closed with #35

PR #35, opened 2015-04-03
Respect rememberSpeed setting

jordn:
Fix for  https://github.com/igrigorik/videospeed/issues/34
Playback was retaining the playback speed between videos even with the
‘remember speed’ setting unchecked.

I'm not convinced this solves the _entire_ issue as there playback speed seems to still be retained on Youtube AJAX loads.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -2,12 +2,13 @@ chrome.extension.sendMessage({}, function(response) {
 
   var tc = {
     settings: {
-      speed: 1.0,         // default 1x
-      speedStep: 0.1,     // default 0.1x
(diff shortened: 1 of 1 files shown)


igrigorik:
Ah, you're right, I lost this logic in earlier refactor -- doh! Thanks for the fix.

Re, YT: I think the player is being reused and hence the controller logic is never called, which results in retained playback speed. I wonder if listening to [loadstart](http://www.w3.org/TR/html5/embedded-content-0.html#mediaevents) event might do the trick.. we can just reset the speed on that.

PR #46, opened 2015-05-24
Update our speed setting when the speed is changed externally.

oggy:
e.g. Coursera videos have their own speed controls. If the speed
is changed from A to B via their controls, then the video is
paused & resumed, our play callback would reset the speed to A.

We now ensure our speed setting is accurate by updating it in the
ratechange callback, which happens no matter how the speed is
updated.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -44,6 +44,9 @@ chrome.extension.sendMessage({}, function(response) {
         });
 
         target.addEventListener('ratechange', function(event) {
+          if (target.readyState === 0) {
+            return;
+          }
           var speed = this.getSpeed();
           this.speedIndicator.textContent = speed;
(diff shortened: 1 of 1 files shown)


igrigorik:
Hmm, this is counter to: https://github.com/igrigorik/videospeed/commit/2e7308c2e9dedb2128e23568a59643158fa75399 - the ratechange logic was causing issues with some other players (if memory serves, Khan Academy..).

oggy:
@igrigorik Interesting. This second change seems to fix the issue for me on YouTube/KA, and feels like a cleaner solution.

igrigorik:
@oggy hmm, can't reproduce the earlier error.. which is good news. Merging, thanks!

PR #49, opened 2015-06-25
Add advance capability; Allow Changes when paused

DalJeanis:
1) Add capability to advance (skip forward) by pressing T.  I would have chosen F for advance, but that resulted in full screen mode.

2) Allow to change speed, rewind and advance when paused.  In testing, I found no issues with allowing the user to adjust the run speed, rewind or advance the video while paused, which seems a useful behavior to me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,11 +15,12 @@ HTML5 video provides a native API to accelerate playback of any video. The probl
 
 #### *[Install Chrome Extension](https://chrome.google.com/webstore/detail/video-speed-controller/nffaoalbilbmmfgbnbgppjihopabppdk)*
 
-Once the extension is installed simply navigate to any page that offers HTML5 video ([example](http://www.youtube.com/watch?v=E9FxNzv1Tr8)), and you'll see a speed indicator in top left corner. Hover over the indicator to reveal the controls to accelerate, slowdown, and quickly rewind the video. Or, even better, simply use your keyboard:
+Once the extension is installed simply navigate to any page that offers HTML5 video ([example](http://www.youtube.com/watch?v=E9FxNzv1Tr8)), and you'll see a speed indicator in top left corner. Hover over the indicator to reveal the controls to accelerate, slowdown, and quickly rewind or advance the video. Or, even better, simply use your keyboard:
 
 * **a** - rewind video 10 seconds.
 * **s** - decrease playback speed.
 * **d** - increase playback speed.
+* **t** - advance video 10 seconds.
 
 Note that you can customize these shortcuts in the extension settings page. Also, a few tips for enabling and forcing HTML5 video:
 
--- a/inject.js
+++ b/inject.js
@@ -5,10 +5,12 @@ chrome.extension.sendMessage({}, function(response) {
       speed: 1.0,          // default 1x
       speedStep: 0.1,      // default 0.1x
       rewindTime: 10,      // default 10s
+      advanceTime: 10,     // default 10s
       resetKeyCode:  82,   // default: R
       rewindKeyCode: 65,   // default: A
       slowerKeyCode: 83,   // default: S
       fasterKeyCode: 68,   // default: D
+      advanceKeyCode: 84,  // default: T
       rememberSpeed: false // default: false
     }
   };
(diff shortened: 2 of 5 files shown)


igrigorik:
@DalJeanis sounds like you're targeting this for YouTube -- correct? Guessing based on your comment for "F" going to fullscreen. If so, are you aware that YT provides arrow key shortcuts already? Also, you can do 1/2/3/... to jump to 10/20/30%... in the video.

DalJeanis:
I was testing on youtube and on vimeo.  Feel free to use something other than T if you prefer.  I was just noting I couldn't use F as default, which felt like the natural choice to me.

DalJeanis:
Is there something I else need to do regarding this?

igrigorik:
Apologies about the delay. 

The code itself looks good, I just keep waffling on whether rewind+advance is something that needs to be part of the extension. I don't have hard usage data on this, but anecdotally (from speaking to those that I know that use the extension) the rewind functionality is rarely used.

That said, I'll go ahead and land this.. :)

DalJeanis:
NP, this is my first contribution to any project on GitHub, so I wasn't sure if I had followed protocol.  I can assure you that *I* use the rewind all the time, when reviewing long training videos I get distracted or interrupted occasionally and have to rewind and figure out what I missed.  Sometimes high speed on a bad accent requires the same.  Having Rewind set to 30-40 seconds, and forward set to 10-15, I can get to the right place pretty quickly.

I was wondering whether I needed to figure out how to update the picture and video and so on.  I hate to leave the docs out of sync with the product.

igrigorik:
@DalJeanis no no, everything on your side was top notch -- thanks a lot for the contribution! :)

I'll take care of the rest (picture, etc) when I push out the next release.

rajington:
If you don't mind I'll make some comments as a huge fan and big user of this extension (literally every video I watch) and I've been trying out local edits about this same topic.

I use rewind heavily, mainly when I mishear something. If I need to fast forward I found myself over-increasing the speed and then resetting to my own speed (e.g. I watch at 2x but set it to 4x then back to 2x when I need to go forward). I'm not sure if this is because of the lack of skip ahead or whether I prefer "catching myself up" in fast forward, but I do it knowing that the arrow keys are there. I like having a custom skip amount (half of YT default) and having it on every video, so since you do have a key that skips backwards (despite the arrow keys) it makes sense that you'd want the interface to supply a skip forward as well.

Generally, I agree on your worry that these advanced capabilities might crowd the design (note that I haven't rebuilt the extension so I'm not 100% sure how it's implemented other than seeing the code changes). At the same time, I see this extension being a huge boon to the power user and really gaining in popularity, so these things should be there. I think a good solution would be to have these hotkeys unset by default, and optionally be able to be visible in the GUI interface.

If you're curious these are the local edits I've made/in the process of making, and I'll make pull requests if you like any of them once I clean them up:

- "Reset speed" toggles between 1.0 and last set speed
- play/pause, same vein as skip ahead, i know space does it on most sites but even on youtube sometimes it scrolls accidentally
- Skip ahead
- Another Finer control option (I have my main at .5 increments and this one at .1) although I really think I'm the only one crazy enough to want this
- Griffin Powermate support (it's hacky though)

igrigorik:
@rajington thanks for the feedback!

re, reset speed: latest release includes "R" shortcut that resets to 1.x. It doesn't remember the last speed though, seems like that'd introduce some gotchas - e.g. I was at 2.x, then I reset and speed up to 1.3x. Then I hit R again, does that go to 2.0 or 1.0? It's not clear what behavior this should take.

Finer rate, griffin: heh, nice. Yeah, don't think this core material :)

DalJeanis:
Play/Pause is the main one of those that I'd vote for.  It would be convenient to have that on the keyboard.  During my testing the advance function it was causing a pause that I had to account for, and it would have been helpful to have at that point.  All better now, though.

rajington:
How I did it was basically a "clean/dirty" state rather than a toggle between two speed settings. So basically in your example it would go back to 1.0 because it was made "dirty" when you increased it to 1.3x. I'm just trying to explain how mine works, not recommending it as an addition necessarily, a lot of these are pretty niche.

On a side note about advance and why I added those "finer control" capabilities is because scrubbing via the extension (HTML 5) is MORE responsive than YouTube's own navigation. Meaning rewind in the extension is a couple of milliseconds faster for me than pushing left arrow. You really notice the difference when you have the video loaded and then you press-and-hold left arrow, vs press-and-hold the rewind key (you can hear while rewinding). I think YouTube either added that delay in so it wouldn't sound garbled when rewinding, so it uses less resources on their end, or some performance/keyup reasoning.

igrigorik:
@rajington ah, the 'dirty bit' approach makes sense, thanks. Re, perf: interesting.. I'll see if I can bug some folks at YT to take a look at this.

PR #57, opened 2015-08-13
Updated READ.me with tip for Wistia vids

joshdance:
Wistia works, just have to refresh the page after you switch to HTML5. I couldn't get it to work for a while, so added to the doc to prevent future pain. 😀


igrigorik:
thanks!

joshdance:
👍

PR #80, opened 2016-01-19
Control videos in iframes

oggy:
This should fix the plugin for videos embedded in iframes, which I seem to encounter quite a bit. Two recent examples are [this embedded Vimeo](http://www.npr.org/sections/krulwich/2012/02/24/147367644/six-legged-giant-finds-secret-hideaway-hides-for-80-years), and [this O'Reilly video](https://player.oreilly.com/videos/9781491933107).

Note this is built on top of #79.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -183,7 +183,7 @@ chrome.extension.sendMessage({}, function(response) {
               iterations: 1,
               delay: 0
             });
-            player.onfinish = function(e) {
+            controllerAnimation.onfinish = function(e) {
               controller.style.visibility = 'hidden';
             }
           }
--- a/inject.js
+++ b/inject.js
@@ -33,162 +33,119 @@ chrome.extension.sendMessage({}, function(response) {
     }
   );
 
+  function defineVideoController() {
+    tc.videoController = function(target) {
+      this.video = target;
+      this.document = target.ownerDocument;
+      if (!tc.settings.rememberSpeed) {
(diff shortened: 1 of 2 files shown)


igrigorik:
I'm having a hard time reading this, because we seem to be mixing code changes with formatting changes. Any chance you can preserve the original formatting / indenting?

oggy:
With the exception of the `tc.settings` stuff at the top I think the only reason formatting changed is because I had to move code around a bit. Does adding `?w=1` to the url when viewing the diff solve the readability?

The main goal in moving code was to extract the video controller definition into a method that's only called when setting up the root document, and not the iframe documents, and then parameterizing this by the document. Note that this part was also done in a separate commit in case that helps grok the changes.

Let me know if it's still hard to follow, and I'll see if I can break it down more. Thanks for taking a look!

igrigorik:
`?w=1` does help. With that in place, I think this looks sane. I've merged it in a local branch and going to try using it locally for a bit to see if I spot any issues out in the wild. 

Interestingly enough, the vimeo link you referenced seems to work for me even without this patch -- the controller doesn't come up for you? 

Slight tangent: now that the controller is 'embed aware', I'm wondering if we could do something smart about routing top-window commands to active embedded player.. E.g. the O'Reilly video is hard to control: you have to click on the video first and then it responds to shortcuts. Hmm.

PR #83, opened 2016-02-09
Added duplicate controls for non latin keyboard layouts.

AKtor:
It was impossible to control the extention when the keyboard layout is not latin.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -2,32 +2,42 @@ chrome.extension.sendMessage({}, function(response) {
 
   var tc = {
     settings: {
-      speed: 1.0,          // default 1x
-      speedStep: 0.1,      // default 0.1x
-      rewindTime: 10,      // default 10s
-      advanceTime: 10,     // default 10s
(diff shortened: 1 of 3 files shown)


igrigorik:
Hmm, is there a reason why these can't be just configured by the user in options, when they install the extension? Defining duplicates keys in code doesn't scale, since we'll end up with this logic for every major language.

AKtor:
It's can't be configured in options, because char-codes of simbols which ara on the same key are not equal in different layouts. For example, char-code of R is 82 and char-code of cyrillic К is 1050.

You are right that my solution is not scalable. Please deny the pull request.

The correct way to use scan-codes of keys instead of char-codes of simbols. So, needs to use keydown event instead of keypress.

igrigorik:
@AKtor ah, makes sense. Switching to keydown seems like the right solution here. Would you be up for taking a crack at it?

AKtor:
Yes I would)

PR #86, opened 2016-02-11
Keypress event changed to keydown. It allows to control the extension with different keyboard layouts

AKtor:
I wonder so few changes. Especially in options.js only one replacement of keypress to keydown and it works well.


igrigorik:
Seems to work as advertised -- thanks! I'll test this locally for a few days and if all goes well I'll push this out in the next release.

PR #87, opened 2016-02-22
Add new hangouts.google.com URL to excluded matches.

### LINKED ISSUE #81: Feature Request: Excluded URLs option
I see the Videospeed over my hangouts video. 
selection_713

Here is the way Vimium does it:
selection_714

ElijahLynn:
Sometimes, but not all the time, I end up at hangouts.google.com instead
of plus.google.com. I think this is part of Google's move to decouple
Google Plus from the Hangouts service.

Resolves #81.


igrigorik:
Good catch, thanks!

Issue #89, opened 2016-03-04
Numpad keys don't work with 0.2.7


igrigorik:
From https://github.com/igrigorik/videospeed/commit/bfb58a749dfc40611819ad38f2c1d3c4cd61d93f#commitcomment-16492427:

> Can't use numpad shortcuts anymore:
> - custom shortcuts stopped working ( + and -)
> - Setting numpad shortcuts shows letters instead of numbers and +-/.

Thanks for reporting this. This is almost certainly due to change in e1864c1. Doing some research..

Mootools has special handling for numpad: https://github.com/mootools/mootools-core/blob/035d7e4547e49963a4422764b2e83232dc60b2cf/Source/Types/DOMEvent.js#L49-L52
Handy test tool http://lawrence.ecorp.net/inet/samples/keycodes.shtml

/cc @ghaith99 @alexkudrow

ghaith99:
Hello Igrigorik,

If I want to get the lastest working 2.6 version before the update to 2.7, which commit should I download, I need it for online lectures, I will install it as Developer until problems hopefully fixed.

Thanks so much

igrigorik:
@ghaith99 you want to checkout at https://github.com/igrigorik/videospeed/commit/20fc3472eb428317861ba75463b1cf8ab3b1bd03. Clone the repo and `git reset --hard 20fc3472eb428317861ba75463b1cf8ab3b1bd03` -- that should do it.

ghaith99:
does this work with videos in all accessible iframes ? the new update works only on video iframe(no shortcuts work when focus is out of video iframe)

alexkudrow:
I didn't expect such behavior. In the handbook, which I used, nothing was written about unusual scan-codes of numpad keys.

I would like to fix it if you don't mind.

igrigorik:
@alexkudrow I dug in a bit but didn't get far.. If you're willing to tackle it, that'd be great!

alexkudrow:
@igrigorik please look at my commit c7d7df2c1d896fdf2f80c460715af62d6c1560cc.

Keypress event triggers only when character key pressed. So, using keydown event needs to process only character keys, and skip others. Moreover, scan-codes only of letter and number keys are correctly matched to keycodes of their symbols. I added aliases for numpad keys, punctuation marks and some other keys to fix this problem. I think this is the easiest way to convert their scan-codes to acceptable form.

Some sources which i used:

[...]

PR #96, opened 2016-03-23
Fixed issue with numpad and punctuation keys

alexkudrow:
Changing keypress event to keydown caused issue #89.

Keypress event triggers only when character key pressed. So, using keydown event needs to process only character keys, and skip others. Moreover, scan-codes only of letter and number keys are correctly matched to keycodes of their symbols. I added aliases for numpad keys, punctuation marks and some other keys to fix this problem. I think this is the easiest way to convert their scan-codes to acceptable form.

Some sources which i used:

- [Javascript keyCodes](https://css-tricks.com/snippets/javascript/javascript-keycodes/)
- [KeyCode.info](https://github.com/wesbos/keycodes)


igrigorik:
:+1:

PR #110, opened 2016-06-13
Check for document before checking for readyState

wesbos:
Every now and then this fires before the document is ready and I end up with this error going nuts in my console: 

Pretty sure this happens because the check happens even before the document exists.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -122,7 +122,7 @@ chrome.extension.sendMessage({}, function(response) {
 
   function initializeWhenReady(document) {
     var readyStateCheckInterval = setInterval(function() {
-      if (document.readyState === 'complete') {
+      if (document && document.readyState === 'complete') {
         clearInterval(readyStateCheckInterval);
         initializeNow(document);
       }


igrigorik:
Hmm, that's an odd one. Can you reproduce this on particular pages?

wesbos:
After some more testing, I did figure it out that it only happens when LastPass inserts the "do you want to remember this password" bar. 

So it's probably more an issue with LastPass?

igrigorik:
It may be, but we should probably handle it here as well.. 

They're injecting an iframe that is removed/goes away? =/ 

A gotcha with proposed change is that we never cancel the setInterval..

wesbos:
Hrm it just happened again without the lastpass extension - so it's not isolated to that.

wesbos:
It just happened again without that extension, but I did get a little bit more info about when it happened:

wesbos:
@igrigorik Some more info - I recorded my screen when this issue popped up. Hopefully there are some insights in here?

Video → http://wes.io/gSTy

igrigorik:
@wesbos can you please try installing the latest extension from git? Clone it and install it as an "unpacked extension"? I installed lastpass and can't seem to reproduce this behavior.. Looking at your video, the error is triggered on a different line from what's currently in master.

igrigorik:
/me nudges @wesbos.. any chance you can try with latest build and see if the problem is still there?

PR #113, opened 2016-06-19
fix NodeList missing foreach in some browsers

OutOfBrain:
`NodeList` does not always have `foreach` method and tends to fail.
Used already existing `forEach` variable to fix.

https://developer.mozilla.org/en-US/docs/Web/API/NodeList


wdhorton:
Didn't realize there was an open PR for this, I just made an issue #114

igrigorik:
thanks for the fix!

Issue #115, opened 2016-06-21
youtube full screen gets offset


TheEpirus:
this happens only when this extension is enabled, when i go into full screen mode the 1-2 cm at the top of the screen goes black and the video is moved down so the 1-2 cm at the bottom of the video gets cut off. this only happens in full screen and i can see the video fine when it is not in full screen

darkkain:
> this only happens in full screen 

Nope. 
On 1920x1080 not full screen have small gap: ![notFullScreenYouTube](http://i.imgur.com/SUo4OGG.jpg) 

And on fullscreen:  ![FullScreenYouTube](http://i.imgur.com/qkZejDH.jpg)

igrigorik:
Resolved via https://github.com/igrigorik/videospeed/pull/116.

PR #116, opened 2016-06-22
Move controller on YT using relative positioning

arao456:
There was a black bar issue on YT reported at #115. This should fix that issue by using `relative` positioning and the `top` style attribute.

image


igrigorik:
👍 ... thanks for working on this!

Issue #82, opened 2016-02-09
Move Controls 


Shot007:
Please make it so that controls can be moved especially on Netflix videos, this is extremely annoying and it makes it difficult to increase/decrease the speeds on Netflix videos.

atlas6523:
This was already opened as this issue: #75 https://github.com/igrigorik/videospeed/issues/75 . It's always a good courtesy to do a little checking to make sure that your issue hasn't already been asked or discussed.  It makes it easier for everyone, including yourself, to see what the group's thoughts are on the problem.

Shot007:
I saw it. But also noticed it was over 2 months ago. I thought a newer post might make the developer see how important this is and how others are also affected by it.

igrigorik:
Closing, landed via https://github.com/igrigorik/videospeed/pull/120.

Issue #75, opened 2015-12-03
Control in the same spot as Netflix control in full screen


meeple142:
Rather than have you try to accommodate for every html 5 video player, hat if you added x and y % location settings so the user can just put it where they want it. So if I wanted it in the top right instead I would set location to something like x = 85%, y = 10%. 

By the way thank you for the awesome add on!

igrigorik:
Ideally, it would be draggable.. My last attempt at that wasn't very productive though. As in, the basics worked, but it wasn't reliable and added a lot of complexity. But, perhaps that's more a comment about my implementation, not the fact that it can't or shouldn't be done :)

atlas6523:
I would think 5 preset orientations would fix this problem for the majority (if not all) of video players.  Upper-left (current position), upper-middle, upper-right, middle-left and middle-right.  I exclude the bottom because that's where the original video player controls are almost always located.  I totally understand the desire to make it drag-able (being an idealistic problem-solver myself), but I think this would be a great in-the-middle solution, at least until you have time to invest into getting it "just right".  Anyways, thanks so much for this great tool.  Hope to hear from you soon.

pjrobertson:
I came over here with the same problem. I agree an option in the preferences would be the simplest and best fix (for now at least). Even just two options of: "Top Left" and "Top Right" would do

igrigorik:
Merging with https://github.com/igrigorik/videospeed/issues/82.

PR #120, opened 2016-06-26
[enhancement] Add ability to drag controller

arao456:
This PR would allow users to drag the controls. The logic is pretty simple, and the position is not saved so it will be reset on page reload. Hopefully this doesn't add too much complexity to the extension, I'm not quite sure of a better way to implement it.

image

#82 #75

### DIFF
--- a/inject.css
+++ b/inject.css
@@ -50,3 +50,7 @@
 div.video-wrapper + div.target {
   height: 0;
 }
+
+.vcs-dragging {
+  cursor: -webkit-grabbing;
+}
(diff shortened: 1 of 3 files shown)


igrigorik:
Awesome! :-)

On a first pass, the implementation looks reasonable.. One thought/question: do we need an extra button? Could we instead make the container itself (the black part, including the speed indicator) as a draggable area?

arao456:
Working on it! There seems to be some issues when the player goes fullscreen, it looks like my offset calculations don't work as intended in that case

PR #142, opened 2016-09-01
fix double controller on Coursera

marshallaf:
Removes and reinitializes controllers if video is re-added to the document.


igrigorik:
👍 thank you!

Issue #69, opened 2015-10-13
"S" key default shortcut brings up "Share" overlay in Vimeo


hazmatte:
Trying to figure out an alternative shortcut key to use instead.  It seems every key I try is mapped to some shortcut in Chrome.  Any "safe" keys that can be used instead of S for slowing down the video?

igrigorik:
Try `Shift-s` :)

PR #168, opened 2017-01-24
add turbo speed for faster speed changing

siadat:
Holding Shift and pressing the increase/decrease keys changes the speed in larger steps. I press Shift+D to quickly skim a video until a point where I think it is interesting. Then I press Shift+S or R.

I felt it natural that holding Shift should increase the step size by which the speed is changed.

The reason I chose the default value of 3.0x is that in Chrome and on Youtube the audio is cut off for playback above 4x (1x+3x).

Note that this will break the workarounds suggested for issues like #69.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -3,6 +3,7 @@ chrome.extension.sendMessage({}, function(response) {
     settings: {
       speed: 1.0,           // default 1x
       speedStep: 0.1,       // default 0.1x
+      turboSpeedStep: 3.0,  // default 3.0x
       rewindTime: 10,       // default 10s
       advanceTime: 10,      // default 10s
       resetKeyCode:  82,    // default: R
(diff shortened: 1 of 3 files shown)


igrigorik:
I like the functionality.. but unfortunately breaking #69 is a non-starter,
I think. We'd need to use a different modifier.

siadat:
I agree. Looking at the modifiers:

- Shift: Conflicts with workaround solution for #69 and others.
- Control: Control+S and Control+D on Chrome conflict with Save and Bookmark shortcuts on Linux. They are available on macOS.
- Alt: Some people use Alt as their modifier for their window manager.
- Fn, Meta, Hyper, OS: Not available on all keyboards.

Another possibility could be to consider using other keys, e.g., Q and E. I would have preferred keys that are right above S and D, but W is already used in Youtube to preview captions.

siadat:
I just came up with an alternative feature that might be worth considering. I'll explain it here, let me know your thoughts.

The problem is that we want to quickly toggle between very fast speed and the default speed. A toggle between current speed and reset speed could resolve this problem. The R key (or a new key) could be used to toggle between the current speed and 1.0x. Here's an example scenario I am thinking of:

- press S multiple times to increase the speed in small 0.10x steps until it is 4.0x
- press R to reset to 1.0x
- [new] press R again to toggle back to the previous non-1.0x speed
- press R again to reset to 1.0x

igrigorik:
Using reset as a quick toggle seems worth exploring: no extra keys, simple. Also, I think it would address https://github.com/igrigorik/videospeed/issues/169.

PR #173, opened 2017-02-01
save and restore toggleSpeed using chrome storage

### LINKED ISSUE #169: Shortcut to a set speed.
It would be nice if there was a new shortcut for a custom speed.

I watch YT at 2.5x most of the time but often listen to music or whatnot and that requires default speed.
My speed step is 0.25x so going between 1x and 2.5x can be a bit annoying. So if, with a new shortcut, I could go straight from 1x to 2.5x it would make things alot more convenient.

Thanks for all the great work on this addon.

siadat:
Fixes #169, again.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -2,6 +2,7 @@ chrome.extension.sendMessage({}, function(response) {
   var tc = {
     settings: {
       speed: 1.0,           // default 1x
+      resetSpeed: 1.0,      // default 1x
       speedStep: 0.1,       // default 0.1x
       rewindTime: 10,       // default 10s
       advanceTime: 10,      // default 10s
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -61,26 +64,27 @@ chrome.extension.sendMessage({}, function(response) {
...
+        var speed = this.video.playbackRate;
+        this.speedIndicator.textContent = speed.toFixed(2);
         tc.settings.speed = speed;
-        chrome.storage.sync.set({'speed': speed});
+        if(speed !== 1.0) {
+          tc.settings.toggleSpeed = speed;
+          chrome.storage.sync.set({'speed': speed, 'toggleSpeed': speed});
+        } else {
+          chrome.storage.sync.set({'speed': speed});
+        }
```
igrigorik: Would it make more sense to only store the reset speed value when reset action is invoked...?

igrigorik:
thanks! 👍

PR #174, opened 2017-02-02
set minimum "saved" speed to 0.0625

### LINKED ISSUE #166: youtube vids wont play - spinning wheels - errors:  "if video doesn't play, restart device"
sent to google, but after disabling every extension i isolated which extension was culprit and further examination while writing this up before clicking submit revealed that speed = 0 was culprit.  but youtube/chrome was giving errors stating:  "if video doesn't play, restart device" with spinning wheels.

issue description:
suddenly cant play youtube vids in chrome except in incognito window, no extension conflicts reported, cleared cache/cookies, restarted chrome often, disabled 'auto play' in youtube and in 'magic actions' and disabled 'magic actions' and restarted, even rebooted windows and still no change.  i guess i'll disable extensions until i discover something, but is there anything else i'm missing?  

solution:
'videospeed controller' was the extension in play - toggling it on/off enables me to play or not play youtube videos.  further examination reveals issue is speed = 0.0 so i recommend the extension informs user in some way when speed is 0.0 to prevent them 5 days of troubleshooting, switching browsers, etc.

canarslan12:
Values below this threshold break some players - closes #166.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -51,6 +51,11 @@ chrome.extension.sendMessage({}, function(response) {
       if (!tc.settings.rememberSpeed) {
         tc.settings.speed = 1.0;
       }
+      else
+      {
+        var s = Math.max(tc.settings.speed, 0.0625);
+        tc.settings.speed = Number(s.toFixed(2));
+      }
       this.initializeControls();
 
       target.addEventListener('play', function(event) {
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -51,6 +51,11 @@ chrome.extension.sendMessage({}, function(response) {
       if (!tc.settings.rememberSpeed) {
         tc.settings.speed = 1.0;
       }
+      else
```
igrigorik: If we want to enforce a minimum on saved speed, we should do so when we save the value: https://github.com/canarslan12/videospeed/blob/94dee5a802dd76739e56898c0b9b3339b621265c/inject.js#L72

igrigorik:
*bump* :)

Issue #158, opened 2016-12-30
Question: Add speed controler for <audio>?


cliffordfajardo:
Today I listening to a podcast online and wanted to increase the playback speed. It was fairly easy to do that after grabbing the `audio` element and setting it's playback rate.

```
var audio = document.getElementById("myAudio");
audio.playbackRate = 0.5;
```

My question is: would it make sense to open a PR & write the code for this? Or is this project focused on only on videos?

igrigorik:
Related discussion: https://github.com/igrigorik/videospeed/issues/122

It would be a relatively simple thing to add. The simple solution for the controller would be simply to hide it for audio - i.e. keyboard shortcuts only. The more involved solution would be to to show the speed on the icon in the toolbar..

alystair:
Oops didn't see this... +1!

canarslan12:
as far as I see, Soundcloud is not using `<audio>` so it's not related with this issue

george-viaud:
I believe this should be a fairly simple thing to add (perhaps just updating the selector pattern) as HTML5 Audio and Video both support the same playbackRate variable if I'm not mistaken.

This is super useful when listening to podcasts online! Right now I have a custom HTML file I use to playback podcasts locally at high speed which is cumbersome - it would be great if this feature was exposed in all online HTML5 Audio I listen to.

Many thanks!

queuebit:
If I stream an `mp3` in Chrome directly from source, it wraps it in a video tag and `videospeed` ends up showing like this.

Just sharing what happens in one specific case:

<img width="366" alt="screen shot 2017-07-31 at 11 47 22 am" src="https://user-images.githubusercontent.com/945262/28786105-35c968ee-75e6-11e7-8cc4-0e50814a6cfc.png">

canarslan12:
Yes it's easy in terms of changing speed but overlay may(almost every time) not shown. There need to be something to show it and make hot keys available. And there's another question, where hotkeys available, on anywhere on site or only focused on "audio" tag. Most of the sites there is no displayable audio element, they are hidden.

George Viaud <notifications@github.com<mailto:notifications@github.com>> şunları yazdı (31 Tem 2017 18:43):


I believe this should be a fairly simple thing to add (perhaps just updating the selector pattern) as HTML5 Audio and Video both support the same playbackRate variable if I'm not mistaken.

[...]

PR #175, opened 2017-02-02
Add speed controler for <audio> - Step 1

canarslan12:
Add speed controler for <audio> #158
NOT FULLY TESTED
Also need more work about controller

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -262,7 +262,7 @@ chrome.extension.sendMessage({}, function(response) {
       });
       observer.observe(document, { childList: true, subtree: true });
 
-      var videoTags = document.getElementsByTagName('video');
+      var videoTags = getVideoElements();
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -323,6 +323,12 @@ chrome.extension.sendMessage({}, function(response) {
       }
     });
   }
+ function getArrayFromTag(tagname) {
```
igrigorik: We can just used querySelector, no need for the extra gymnastics.

That said, handling audio requires a much deeper rethink of how the plugin works and how the speed is displayed and where. I don't think we should be landing this half-baked.. as it opens up a deep well of technical debt.

igrigorik:
Closing, as this is a partial solution that would open up a lot of technical debt.. If/when we want to tackle audio as a first-class use case, we can revisit.

PR #185, opened 2017-03-09
display saved time

siadat:
I wrote this as an experiment after a few trials and errors. Here's a screenshot:

screenshot_2017-03-09_13-44-10

Saved time is

    (total length of the video that is watched) - (time spent to watch it)

Examples:
- If speed is 1x, then 0s is saved
- If speed is 2x and video is 10m, then 5m is saved
- If speed is 1x for 5m and 2x for 5m, then 2m30s is saved
- If speed is 2x and video is 10m and user watches it twice, then 10m is saved

I had to calculate the time that user spends watching the video playing. I use the `playing`, `pause`, and `waiting` events. The `waiting` event has the same effect as the `pause` event here. For example, it is triggered when the video is playing but the user is waiting for the buffer to catch up and load. During this time nothing is played for the user.

Let me know if I should change anything. Thanks!

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -56,6 +56,79 @@ chrome.extension.sendMessage({}, function(response) {
       }
       this.initializeControls();
 
+      let calcDurations = function(target) {
+        let from = Number(target.dataset['playFrom']);
+        let to = Number(target.dataset['playTo']);
+        let rate = Number(target.dataset['playRate']);
+        let spentDuration = Number(target.dataset['spentDuration']) || 0;
+        let playedDuration = Number(target.dataset['playedDuration']) || 0;
+        if (target.dataset['playStatus'] === "playing") {
+          to = Date.now();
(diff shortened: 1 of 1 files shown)


igrigorik:
This is neat! However, I'm not sure if we actually want to land this...

- It's a fun FYI, but I suspect most people will want to hide it by default
- Providing an opt-in setting for it is additional complexity + ongoing maintenance burden
  - It adds a sizable piece of code to maintain moving forward as is; plus opt-in.

siadat:
I understand! Thank you for your comment and for the explanation! :)

marketb12:
I agree with the other comments. Adding it to the widget would add some unneeded bloat. Having it on the menu page would be cool though.

siadat:
Thank you for your feedback. I agree. I will have a go at implementing it then I will update this PR to discuss it further.

PR #188, opened 2017-03-16
use mouse wheel scroll to increase/decrease speed

metaist:
After a few months of use, I realized that despite adjusting the speed frequently, my natural instinct is to use the mouse rather than the keyboard (not sure why; I use keyboard shortcuts all the time).

Here's a little patch that allows you to use the mouse wheel anywhere on the overlay to increase / decrease the speed. Originally, I had it on the `.draggable`, but then I bumped it up to the `#controller` to take advantage of [Fitt's Law][wiki-1] (small increases in target area of a small thing matter more than comparable increases on a large hit area).

[wiki-1]: https://en.wikipedia.org/wiki/Fitts%27s_law

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -121,6 +121,13 @@ chrome.extension.sendMessage({}, function(response) {
         </div>
       `;
       shadow.innerHTML = shadowTemplate;
+      shadow.querySelector('#controller').addEventListener('wheel', (e) => {
+        runAction(e.wheelDelta > 0 ? 'faster' : 'slower', document);
+        e.preventDefault();
+        e.stopPropagation();
+      });
+
+
       shadow.querySelector('.draggable').addEventListener('mousedown', (e) => {
         runAction(e.target.dataset['action'], document);
       });


igrigorik:
I just realized that I don't own a single mouse with a wheel.. :)

I do wonder how many people will ever find this, but overall it seems like a good (hidden) feature. Thanks for the patch!

metaist:
Good point on the low discoverability; not sure why that was my first instinct when I installed the extension.

strabbit:
I discovered this change after a couple of days of trying to figure out why YouTube videos would (what seemed like) randomly play back at 0.06x speed. I had never actively adjusted the playback speed that low (nor would I ever).

It turns out that as I was scrolling down my Facebook feed with my cursor at x-coordinates that line up with `#controller`. Even thought I never actively hovered `#controller`, when `#controller` fell beneath my cursor (because I was scrolling down the page), it caused the playback speed to be adjusted to the minimum. It does this for non-playing videos in my feed. I rarely play videos on Facebook, so when I finally noticed that the speed setting had been messed with on another site (YouTube), it was hard to figure out that this was actually was causing the issue. It just felt like the extension was broken and randomly resetting my speed to 0.06x.

For this reason, I plead that this PR be reverted, and if it is reapplied that it be tied to a configuration flag that is off by default. (Which, could also help with the discoverability issue)

igrigorik:
First off, great detective work on tracking it back to the PR.. :)

I agree that this is an unfortunate and unexpected side-effect. @metaist wdyt? I'm generally wary and skeptical of adding more and more options..

metaist:
Ah! Sorry, @strabbit. I, too, am impressed that you tracked that down. And thank you for the additional details about how exactly it was getting reset.

I'm okay with this being rolled back. While I've have exclusively used the scroll wheel, I can see that such a hard-to-detect behavior can be frustrating.

I also agree that simply adding more and more options is is often used to simply offload design decisions to other people.

So unless lots of people are using the scroll wheel, I'm in favor of reverting. Perhaps I'll practice using the keyboard shortcuts.

igrigorik:
@strabbit sg, thanks! Reverted and this will go out in the next release.

strabbit:
Thanks

bryanshannon:
Nevermind my prior comments. This is good stuff!

PR #200, opened 2017-04-11
speed inc. from 0.06 to 0.10 not to 0.16, fixes #198

### LINKED ISSUE #198: Incorrect speed increasing
After decreasing playback rate to minimal (0.06) it is impossible to return back to round values. My inner perfectionist is saddened by it.

Steps to reproduce:
1. Hit "slower" button until minimal (0.06) playback rate is reached
2. Try to set 1.0 rate again by hitting "faster" button multiple times.

Observed behaviour:
Playback rate takes the following values: 0.16, ... , 0.86, 0.96, 1.06, ...

Expected behaviour:
Playback rate takes the following values: 0.1, ... , 0.8, 0.9, 1.0, ...

xftroxgpx:
#198

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -296,7 +296,7 @@ chrome.extension.sendMessage({}, function(response) {
         } else if (action === 'faster') {
           // Maximum playback speed in Chrome is set to 16:
           // https://cs.chromium.org/chromium/src/media/blink/webmediaplayer_impl.cc?l=103
-          var s = Math.min(v.playbackRate + tc.settings.speedStep, 16);
+          var s = Math.min( (v.playbackRate < 0.1 ? 0.0 : v.playbackRate) + tc.settings.speedStep, 16);
           v.playbackRate = Number(s.toFixed(2));
         } else if (action === 'slower') {
           // Audio playback is cut at 0.05:


igrigorik:
Great stuff, thank you!

xftroxgpx:
I am not happy with this PR I made :)) because for some reason I didn't realize that the 0.10 step is user settable even though I've modified the line referencing it(`tc.settings.speedStep`).
I shall attempt a new PR and add in this one too: https://github.com/igrigorik/videospeed/pull/203#issuecomment-294407198

xftroxgpx:
This got complicated fast, I doubt this would be acceptable so I'm not going to PR it  :)
But here it is anyway:

There are 2 unrelated fixed there too (audio playback cutoff typo; and storage speedstep got saved&restored with full digits not toFixed(2) so inc/dec would be +0.12345 not +0.12)

```diff
diff --git a/inject.js b/inject.js
index 58996e0..ad9af62 100644
--- a/inject.js
+++ b/inject.js
@@ -303,15 +303,27 @@ chrome.extension.sendMessage({}, function(response) {
         } else if (action === 'faster') {
           // Maximum playback speed in Chrome is set to 16:
           // https://cs.chromium.org/chromium/src/media/blink/webmediaplayer_impl.cc?l=103
-          var s = Math.min( (v.playbackRate < 0.1 ? 0.0 : v.playbackRate) + tc.settings.speedStep, 16);
-          v.playbackRate = Number(s.toFixed(2));
+          //toFixed(4) because 0.0625, even though it should be toFixed(2) because speedstep is also limited to 2, eg. 0.10
+          var s = Number((Math.floor(
+            Number((v.playbackRate + tc.settings.speedStep).toFixed(4)) / tc.settings.speedStep)
+                * tc.settings.speedStep).toFixed(4));
+          //gets stuck sometimes due to rounding! workaround
+          s = Math.min(s == v.playbackRate ? Number(s + tc.settings.speedStep).toFixed(4): s, 16);
+          console.log("playnow=",v.playbackRate, " step=", tc.settings.speedStep, " nextspeed=",s);
+          v.playbackRate = Number(s.toFixed(4));
         } else if (action === 'slower') {
-          // Audio playback is cut at 0.05:
+          // Audio playback is cut at 0.5:
           // https://cs.chromium.org/chromium/src/media/filters/audio_renderer_algorithm.cc?l=49
           // Video min rate is 0.0625:
           // https://cs.chromium.org/chromium/src/media/blink/webmediaplayer_impl.cc?l=102
-          var s = Math.max(v.playbackRate - tc.settings.speedStep, 0.0625);
-          v.playbackRate = Number(s.toFixed(2));
+          // Compute next slower playback speed that is divisible by speedStep:
+          var s = Number((Math.ceil(
+            Number((v.playbackRate - tc.settings.speedStep).toFixed(4)) / tc.settings.speedStep)
+                * tc.settings.speedStep).toFixed(4));
+          //gets stuck sometimes due to rounding! workaround
+          s = Math.max(s == v.playbackRate ? Number(s - tc.settings.speedStep).toFixed(4): s, 0.0625);
+          console.log("playnow=",v.playbackRate, " step=", tc.settings.speedStep, " nextspeed=",s);
+          v.playbackRate = Number(s.toFixed(4));
         } else if (action === 'reset') {
           if(v.playbackRate === 1.0) {
             v.playbackRate = tc.settings.resetSpeed;
diff --git a/options.js b/options.js
index ab7af96..b386961 100644
--- a/options.js
+++ b/options.js
@@ -102,7 +102,8 @@ function save_options() {
   var startHidden = document.getElementById('startHidden').checked;
   var blacklist     = document.getElementById('blacklist').value;
 
-  speedStep     = isNaN(speedStep) ? tcDefaults.speedStep : Number(speedStep);
+  speedStep     = isNaN(speedStep) ? tcDefaults.speedStep : Number(Number(speedStep).toFixed(2));
+  document.getElementById('speedStep').value = speedStep;
   rewindTime    = isNaN(rewindTime) ? tcDefaults.rewindTime : Number(rewindTime);
   advanceTime   = isNaN(advanceTime) ? tcDefaults.advanceTime : Number(advanceTime);
   resetKeyCode  = isNaN(resetKeyCode) ? tcDefaults.resetKeyCode : resetKeyCode;

```

igrigorik:
Hmm, yeah... "well, that escalated quickly!" ;-)

I do like the approach of ensuring that the increment is divisible by speed step. If we we're to raise the min from 0.0625 to a higher (e.g. 0.1) value, would it make our life easier here?

xftroxgpx:
>  If we we're to raise the min from 0.0625 to a higher (e.g. 0.1) value, would it make our life easier here?

**tl;dr:** easier for the default speedStep of 0.10, yes; but not for any other non-0.1-divisible speedSteps.

Apparently easier but not really(ok, it's easier, for the default speedStep). Depends on decrement step. If speedStep is 0.12 for example, then next increment speed from 0.10 should be 0.12 not 0.22; OR, if fixing current speed to be divisible first, then next speed would be 0.24 which defies speedStep(in the sense that's bigger than it(0.14 > 0.12); lesser is okay though(0.02)) which is why I haven't considered doing that(that=which I also mentioned in the prev. comment's edits).

But it's easier for the default speedStep (of 0.1) yes, definitely! Because it's such a clean divisible speedStep, very nice and cuddly :D But personally, I wouldn't be comfortable losing the lowest allowable speed(0.0625) because of that. The current PR(as it already is merged) is a much better alternative.

I am using the above patch + some other changes on my fork and I'm happy with it, but no way it will even be considered for merger here (too messy & stuff) so I kinda gave up finding a solution for this original repo., at least one that I wouldn't regret doing lol

igrigorik:
Right, OK. I guess if the offset bothers someone.. they can "R" to reset to 1.0 and restart the process. Let's stick with what we have here, for now at least.

queuebit:
@igrigorik Thanks for keeping the core strong and saying no when it makes sense.

PR #202, opened 2017-04-12
keep per tab speed consistent

xftroxgpx:
...in the current browser session,
by saving it before navigating away.

Test:
1. open two tabs with videos,
2. set different speeds for each,
3. reload previous (or any) tab.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -56,6 +56,18 @@ chrome.extension.sendMessage({}, function(response) {
       }
       this.initializeControls();
 
+      //Save the speed of current tab's video before leaving the page
+      //(ie. reloading page or going Backward/Forward in history)
+      //so that it doesn't restore the speed set by a video in another tab
+      //but instead, keeps the current tab's video speed the same,
+      //until changed by user or browser quits.
+      window.onbeforeunload = function(e) {
+        if (target.readyState === 0) {
+          return;
+        }
+        chrome.storage.sync.set({'speed': this.getSpeed()});
+      }.bind(this);
+
       target.addEventListener('play', function(event) {
         target.playbackRate = tc.settings.speed;
       });


inject.js
```diff
@@ -56,6 +56,18 @@ chrome.extension.sendMessage({}, function(response) {
       }
       this.initializeControls();
 
+      //Save the speed of current tab's video before leaving the page
+      //(ie. reloading page or going Backward/Forward in history)
+      //so that it doesn't restore the speed set by a video in another tab
+      //but instead, keeps the current tab's video speed the same,
+      //until changed by user or browser quits.
+      window.onbeforeunload = function(e) {
```
igrigorik: Hmm, this one is tricky.. this behavior you're proposing makes sense for reload and back/forward, but not for new tabs, where the behavior would now be based on tab visibility order (which seems even worse) instead of actually making speed changes.

xftroxgpx: A new tab gets the last seen/fiddled speed, which is either the speed of the last closed/reloaded/Back/Forwarded video tab, or the speed that was last changed eg. via A/D keys, whichever was last.
It seems kinda normal to me. Unless I'm missing something? :)

The only problem is: you don't know which speed will be saved when closing the whole browser or window(with a bunch of video tabs); tab visibility order, I guess, would be deciding which speed was last and that one will be remembered/restored when browser is started again or new video tab opened.


Previously(currently), the speed of the new tab(and on browser restart) was the speed that was last changed eg. via A/D keys.



But, you don't have to consider merging this; I just wanted it to be available here in PRs for anyone else (like me) who needed this behavior, if any.

igrigorik:
> A new tab gets the last seen/fiddled speed, which is either the speed of the last closed/reloaded/Back/Forwarded video tab, or the speed that was last changed eg. via A/D keys, whichever was last. It seems kinda normal to me. Unless I'm missing something? :)

It's common to have videos for ads / short clips nowadays. The 'save last tab' behavior would mean that whenever I switch to a tab with such element and close it — e.g. I'm flipping through tabs while I have a video playing in another tab I'm controlling — it would update the saved speed with respect to last viewed tab.. which is odd, because I probably never interacted with the videos on that page. 

My hunch is that this behavior would cause a lot of confusion and a feeling that "remember is broken". I think sticking with "update default on explicit action" is a safer bet.

> But, you don't have to consider merging this; I just wanted it to be available here in PRs for anyone else (like me) who needed this behavior, if any.

sg. I'll close this with no action, but it'll be available in the history. If there is demand for this behavior in the future, we can revisit. 

cheers!

PR #203, opened 2017-04-16
Playback at "fast" speed

jbrains:
By pressing "G", the viewer can play the video at a "fast" speed. If people like this, then we could give them a configuration setting to choose their own "fast" speed.

I wanted this because I was watching a bunch of online training videos and I wanted to just jump to 2.5x speed for every new video in the course.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -334,7 +334,7 @@ chrome.extension.sendMessage({}, function(response) {
 
   // CONTRACT: video has a way to set the playback rate.
   function playVideoAtFastSpeed(video) {
-    video.playbackRate = 2.5;
+    video.playbackRate = tc.settings.fastSpeed;
   }
 
  function handleDrag(video, controller) {
--- a/inject.js
+++ b/inject.js
@@ -4,7 +4,7 @@ chrome.extension.sendMessage({}, function(response) {
       speed: 1.0,           // default 1x
       resetSpeed: 1.0,      // default 1x
       speedStep: 0.1,       // default 0.1x
-      fastSpeed: 1.75,      // default 1.75x
+      fastSpeed: 1.8,       // default 1.8x
       rewindTime: 10,       // default 10s
       advanceTime: 10,      // default 10s
       resetKeyCode:  82,    // default: R
--- a/inject.js
+++ b/inject.js
@@ -332,8 +332,9 @@ chrome.extension.sendMessage({}, function(response) {
     });
   }
 
-  function playVideoAtFastSpeed(videoDomElement) {
-    console.log("playVideoAtFastSpeed", videoDomElement);
+  // CONTRACT: video has a way to set the playback rate.
+  function playVideoAtFastSpeed(video) {
+    video.playbackRate = 2.5;
   }
 
  function handleDrag(video, controller) {
--- a/inject.js
+++ b/inject.js
@@ -4,6 +4,7 @@ chrome.extension.sendMessage({}, function(response) {
       speed: 1.0,           // default 1x
       resetSpeed: 1.0,      // default 1x
       speedStep: 0.1,       // default 0.1x
+      fastSpeed: 2.5,       // default 2.5x
       rewindTime: 10,       // default 10s
       advanceTime: 10,      // default 10s
       resetKeyCode:  82,    // default: R
@@ -12,6 +13,7 @@ chrome.extension.sendMessage({}, function(response) {
       rewindKeyCode: 90,    // default: Z
       advanceKeyCode: 88,   // default: X
       displayKeyCode: 86,   // default: V
+      playAtFastSpeedKeyCode: 70,    // default: F
       rememberSpeed: false, // default: false
       startHidden: false,   // default: false
       blacklist: `
--- a/inject.js
+++ b/inject.js
@@ -13,7 +13,7 @@ chrome.extension.sendMessage({}, function(response) {
       rewindKeyCode: 90,    // default: Z
       advanceKeyCode: 88,   // default: X
       displayKeyCode: 86,   // default: V
-      playAtFastSpeedKeyCode: 70,    // default: F
+      playAtFastSpeedKeyCode: 71,    // default: G
       rememberSpeed: false, // default: false
       startHidden: false,   // default: false
       blacklist: `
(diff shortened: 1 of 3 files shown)


xftroxgpx:
view without whitespace: https://github.com/igrigorik/videospeed/pull/203/files?w=1

jbrains:
@xftroxgpx Yes. I chose to err on the side of keeping those comments lined up, since I didn't know how much you folks cared about it. :)

igrigorik:
> I wanted this because I was watching a bunch of online training videos and I wanted to just jump to 2.5x speed for every new video in the course.

Curious, why does "remember speed" setting fall short for this use case? If you have that enabled it should store the last playback speed and automatically start the next one at same speed.

jbrains:
@igrigorik I suppose I think of it like a "preset speed", like a favorite station on the radio. I suppose I could use "remember speed" for this purpose, but if I want to go faster or slower on this one video, it's nice to quickly jump back to "preferred fast speed" instead of having to adjust.

igrigorik:
Fair enough, this is a feature that's been requested a few times in the past..

As an aside, I'm wary of adding more and more custom shortcuts -- this process doesn't scale. One escape hatch I can see here is that we can provide a "create your own shortcut" for power users such as yourself: expose an interface to define a key and a command that should be run. This way we can get out of business of writing more and more one-offs and instead let the power users do whatever they need it to do. That said, in this particular case, I thin we can make an exception. I have not personally felt the need for this, but I've heard the request enough times from others..

inject.js
```diff
@@ -1,19 +1,21 @@
...
+      fastSpeed: 2.5,                // default 2.5x
+      rewindTime: 10,                // default 10s
+      advanceTime: 10,               // default 10s
+      resetKeyCode:  82,             // default: R
+      slowerKeyCode: 83,             // default: S
+      fasterKeyCode: 68,             // default: D
+      rewindKeyCode: 90,             // default: Z
+      advanceKeyCode: 88,            // default: X
+      displayKeyCode: 86,            // default: V
+      playAtFastSpeedKeyCode: 71,    // default: G
```
igrigorik: Let's do `fastKeyCode` to be consistent with the rest.

inject.js
```diff
@@ -1,19 +1,21 @@
...
-      fasterKeyCode: 68,    // default: D
-      rewindKeyCode: 90,    // default: Z
-      advanceKeyCode: 88,   // default: X
-      displayKeyCode: 86,   // default: V
-      rememberSpeed: false, // default: false
-      startHidden: false,   // default: false
+      speed: 1.0,                    // default 1x
+      resetSpeed: 1.0,               // default 1x
+      speedStep: 0.1,                // default 0.1x
+      fastSpeed: 2.5,                // default 2.5x
```
igrigorik: As a default, I think most folks will find this too fast. I recommend we set 1.75 as the default, which is the recommended "comfortable" speed for most folks based on past research.

jbrains:
@igrigorik I understand. I think one could extract it quite naturally from what's already there; I imagine it's just waiting for someone really eager to do it. :)

Thanks.

inject.js
```diff
@@ -321,11 +325,18 @@ chrome.extension.sendMessage({}, function(response) {
...
         } else if (action === 'drag') {
           handleDrag(v, controller);
+        } else if (action === 'fast') {
+          playVideoAtFastSpeed(v);
         }
       }
     });
   }
 
+  // CONTRACT: video has a way to set the playback rate.
```
igrigorik: What do you mean by this? All `video` elements can do this..

jbrains: Just documenting the part of `video` that that function cares about. My convention. I'm happy to remove it.

inject.js
jbrains: When I add the configuration ability, I'll change this to 1.75.

inject.js
igrigorik: Gotcha. Yeah, let's drop it, it's implicit here and stands out as a red flag with respect to the rest of the file.

jbrains:
I think I've made the changes you'd asked for, so it's ready for your next review.

xftroxgpx:
One thing: if you set it to 1.75, then you can't get back to 1.00 (since the decrements are by 0.10). Maybe someone(me?) should make a new PR to ensure that the next inc/dec is to a rounded value (but still no more than 0.10 per step)? then 1.75 would be great here(next would be 1.80 then 1.90 or prev. would be 1.70 then 1.60), and it would also handle this case https://github.com/igrigorik/videospeed/pull/200

inject.js
```diff
@@ -1,19 +1,21 @@
...
-      advanceTime: 10,      // default 10s
-      resetKeyCode:  82,    // default: R
-      slowerKeyCode: 83,    // default: S
-      fasterKeyCode: 68,    // default: D
-      rewindKeyCode: 90,    // default: Z
-      advanceKeyCode: 88,   // default: X
-      displayKeyCode: 86,   // default: V
-      rememberSpeed: false, // default: false
-      startHidden: false,   // default: false
+      speed: 1.0,                    // default 1x
```
igrigorik: Nitpick: can you please reverse the whitespace change to original?

options.html
```diff
@@ -33,6 +33,10 @@
         <input id="fasterKeyInput" placeholder="press a key" type="text" value=""/>
       </div>
       <div class="row">
+        <label for="fastKeyInput">Jump to Preferred Fast speed</label>
```
igrigorik: Let's simplify this to "Preferred speed"

options.html
```diff
@@ -57,6 +61,10 @@
           <input id="speedStep" type="text" value=""/>
       </div>
       <div class="row">
+          <label for="fastSpeed">Preferred Fast Playback Speed (x)</label>
```
igrigorik: Ditto, "Preferred Speed". 

(and I just realized we flip casing between shortcuts and settings.. I'll fix this later)

igrigorik:
@xftroxgpx @jbrains I'm good with 1.8 as default.

inject.js
jbrains: Easily enough.

jbrains:
I think I got everything. It's possible that @igrigorik meant something slightly different with the configuration settings labels, but otherwise I do believe I got it all.

igrigorik:
Looks great, thanks for your patience! Merging.. :-)

PR #211, opened 2017-05-05
Revert "use mouse wheel scroll to increase/decrease speed"

igrigorik:
Reverts igrigorik/videospeed#188


bryanshannon:
For what it's worth, I was thinking of adding a very similar feature in a future pull request. Can you please describe why this was backed out? Right up my alley!

ChadBailey:
@bryanshannon I commented on this in issue #581 if you'd like to bring the discussion there (so it isn't spread across so many places)

PR #212, opened 2017-05-06
add small delay to scroll to adjust speed

smokeybears:
I was experiencing the same issue @strabbit described in #188 but on Reddit. Something like this could be used to get the best of both worlds. It adds a small delay to adding the `wheel` event listener, this way it doesn't get accidentally triggered when scrolling down a page but still allows for the functionality. 

I used the scroll to change speed fairly often (once I discovered it) but would also be okay with the feature disappearing, thought I might as well run this by you all the same in case others were enjoying the feature.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -125,6 +125,23 @@ chrome.extension.sendMessage({}, function(response) {
         </div>
       `;
       shadow.innerHTML = shadowTemplate;
+      shadow.querySelector('#controller').addEventListener('mouseover', (e) => {
+        let adjustSpeedOnScroll = (e) => {
+          runAction(e.wheelDelta > 0 ? 'faster' : 'slower', document);
+          e.preventDefault();
+          e.stopPropagation();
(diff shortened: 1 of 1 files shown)


igrigorik:
/cc @strabbit @metaist for feedback.

I think it's an improvement on what we had before, but I'm not sure the added complexity warrants the feature: some will want a shorter delay than others, some will still get caught by surprise, etc. My suggestion is to nudge more folks towards keyboard shortcuts.. :-)

smokeybears:
I agree with that @igrigorik, adding a delay doesn't help discoverability and the feature would probably go unused by most. Also after using the keyboard shortcuts a little bit it's growing on me as a better interface to adjust the speed. Thanks for creating this, it's a super helpful tool 😄

igrigorik:
@smokeybears thanks, closing this. 

p.s. glad you found it useful! :)

Issue #216, opened 2017-05-23
Reset speed use preferred speed at first


canarslan12:
> would be cool.
> I don't like starting videos with high playback rate but I would like to get previous speed that I have with pressing R key once.
> fx: video1 in youtube; pressed d 10 times to reach 4.00 closed it
> video2 opened and pressed r makes speed 4.00

edit: reset speed might use preferred video speed at first, it will be better.

igrigorik:
Isn't that what "remember speed" already accomplishes?

canarslan12:
But reset button is not working at first and I always clicking it.

igrigorik:
As in, you expect reset key to switch to preferred speed? That's not the intent.. or I don't follow.

canarslan12:
I checked code and did it with only changing 1 line. Pulling after some test

igrigorik:
Resolved via #216.

PR #218, opened 2017-05-24
suggestion #216 - Reset use fastSpeed initially

canarslan12:
#216


igrigorik:
Ah, OK, now I follow you -- thanks for tracking it down.

Issue #214, opened 2017-05-19
Not always showing increased speed when using 'Preferred Speed' hotkey


baz357:
Hi,

I've always used the d and s keys to increase/decrease speed. Recently I started to use the 'Preferred Speed' hotkey to directly jump to the 1.60 speed I mostly use, and I found that while it always increases the speed, sometimes it shows the speed as still being 1.00. When I then increase or decrease the speed, it correctly shows the changed value.

Not a big issue and I didn't find a way to reproduce it, just wanted to let you know. Thanks for this lovely extension!

canarslan12:
It might be related with not ready(unset) videos. Can you test it?

baz357:
Yes, you are right - it *is* related. I can sometimes reproduce it by using the 'Preferred Speed' hotkey right after the video loads, if I wait a few seconds it displays the speed correctly.

canarslan12:
I think problem is here. Let's ask why @igrigorik add this code. Is it safe to remove?
'
        if (target.readyState === 0) {
          return;
        }
'

canarslan12:
I checked file history and contributor for that line was @oggy #e75f4f0
he was saying that "Fixes speed setting when jumping between YouTube videos." but there is no problem about it when removing this lines currently

igrigorik:
Closed via #214.

PR #220, opened 2017-05-26
Not updating speed correctly #214

canarslan12:
I tested for several times with/without throttling in different videos and nothing effected/no console error and fixed issue.
edit: wait do not pull for now
edit: that was my mistake, you can pull it
edit: causing loop on tc.settings.resetSpeed and breaks r and g buttons functionality
edit: I can not reproduce with my latest build but I reproduced it with live built. It means safe to pull

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -65,9 +65,6 @@ chrome.extension.sendMessage({}, function(response) {
       });
 
       target.addEventListener('ratechange', function(event) {
-        if (target.readyState === 0) {
-          return;
-        }
         var speed = this.getSpeed();
         this.speedIndicator.textContent = speed;
         tc.settings.speed = speed;


igrigorik:
See: https://github.com/igrigorik/videospeed/commit/e75f4f0c6059c28093c07265e0bdee89f1593549

> Don't update our video speed setting before the video is loaded.
> In this state, the playback rate is always 1, as we haven't had a chance
> to update it yet (which currently happens in the play callback).
> 
> Fixes speed setting when jumping between YouTube videos.

Landing this would re-introduce the above issue.

Can you elaborate which "not updating speed correctly" case you're trying to address?

igrigorik:
Ah, this is for https://github.com/igrigorik/videospeed/issues/214.

I guess we can try it and see if this re-surfaces any old(er) issues..

Issue #221, opened 2017-05-26
resetSpeed and fastSpeed conflict


canarslan12:
when speed is 1.00 and G pressed twice
new speed is 1.00 but resetSpeed is also 1.00 now which is not correct.

same for resetSpeed
when speed is 1.80 and R pressed twice
new speed is 1.80 but resetSpeed is also 1.80 now which is not correct.

igrigorik:
Closing, let's continue in #222.

PR #222, opened 2017-05-26
resetSpeed and fastSpeed conflict #221

canarslan12:


inject.js
```diff
@@ -330,7 +330,14 @@ chrome.extension.sendMessage({}, function(response) {
...
     if (v.playbackRate === target) {
-      v.playbackRate = tc.settings.resetSpeed;
+      if(v.playbackRate === tc.settings.resetSpeed)
+      {
+        v.playbackRate = (target!==1.0)? 1.0 : tc.settings.fastSpeed;
+      }
+      else
+      {
+        v.playbackRate = tc.settings.resetSpeed;
+      }
```
igrigorik: Can you explain what you're trying to accomplish?

canarslan12:
on #221 i explained it

> when speed is 1.00 and G pressed twice
new speed is 1.00 but resetSpeed is also 1.00 now which is not correct.

same for resetSpeed
when speed is 1.80 and R pressed twice
new speed is 1.80 but resetSpeed is also 1.80 now which is not correct

igrigorik:
> when speed is 1.00 and G pressed twice
> new speed is 1.00 but resetSpeed is also 1.00 now which is not correct.

Why does this matter, if hitting G will take you back to correct value anyway?

I'm struggling to make sense of the new codepath you're proposing here. Is there a way to refactor or simplify it, to make it more human friendly? :)

xftroxgpx:
I've found only one reason why that would matter, and that's when you press 'G'/'R' in the context of this issue https://github.com/igrigorik/videospeed/issues/230

canarslan12:
When I use R to change video speed over and over again(1.00-2.00) it's fine.
After that, when I fast forward with G(4.00) when speed is 1.00 and turn back to 1.00 with G again; system forget Reset video speed(2.00) and I should go this speed manually, R does not change anything. 
This is fixing it.

PR #224, opened 2017-06-02
Fix link to media.mit.edu article

SeanSilke:
From broken link
http://xenia.media.mit.edu/~barons/html/avios92.html#beasleyalteredspeech
To new link with the same content
http://alumni.media.mit.edu/~barons/html/avios92.html#beasleyalteredspeech


igrigorik:
Good catch. thank you!

PR #246, opened 2017-07-24
Use onreadychangestate event for script initialization

johnp:
Due to an [issue](https://github.com/codebicycle/videospeed/issues/4) with the Firefox fork of this extension I've had a look at this and it would probably be nicer to use the event instead of constantly polling.
This is untested, but literally a copy-paste from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState) so it should work. The [`load`](https://developer.mozilla.org/en-US/docs/Web/Events/load) event would be semantically equivalent.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -186,17 +186,22 @@ chrome.extension.sendMessage({}, function(response) {
     if (blacklisted)
       return;
 
-    var readyStateCheckInterval = setInterval(function() {
-      if (document && document.readyState === 'complete') {
-        clearInterval(readyStateCheckInterval);
+    window.onload = () => initializeNow(document);
+    if (document) {
+      if (document.readyState === "complete") {
         initializeNow(document);
+      } else {
+        document.onreadystatechange = () => {
+          if (document.readyState === "complete") {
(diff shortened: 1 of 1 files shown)


igrigorik:
In theory this should work.. In practice, it doesn't in many cases. :(

Take a look at the commit history for that code.

johnp:
I checked the commit history back until "initial import" or so and couldn't find anything, but if you say so...
Is the `load` event any better?
Could it be that the event listener gets attached after the event has already fired and adding a manual check for readyState before that helps?

igrigorik:
@johnp thanks for digging into this. Merged your patch locally, going to try to browse around with it and see if I run into any issues.

johnp:
I've changed the ordering a bit (add the window.onload event handler first because `document` may be null if injected early), added a missing null-check for ` document` and squashed the commits. WFM in the Firefox port on Nightly.

igrigorik:
Been running with the earlier version for a while -- everything seems a-ok. 

Merging, thanks for your help and patience! :)

Issue #233, opened 2017-07-03
Not saving speed on netflix


arikroc:
Just noticed this started happening in the past day or two.

After changing episodes using the episode selector on netflix, it appears to reset the speed

Using latest version of chrome stable channel on MacOS

igrigorik:
Please try restarting Chrome, we released new version of the plugin a few days ago.. a clean restart often cures most of these. If not, check your options for the checked flag, and reopen if problem persists.

arikroc:
@igrigorik Can you re-open? I made sure the option was checked (it was), then restarted chrome twice, and it's still an issue. Chrome extensions page lists version 0.4.8

Steps to repro: set speed on netflix, and then refresh the page - it'll go back to 1.00

igrigorik:
sg, let's dig a bit deeper here..

igrigorik:
Huh, interesting.. The extension saves and initializes the prefs correctly, the issue is that netflix is doing something really odd in the unload step of the page and overrides the video speed to "1.0", which then gets persisted and hence the behavior.

igrigorik:
Ok, think I have a patch that fixes it.. 

```diff
diff --git a/inject.js b/inject.js
index 7c5dac3..07881a9 100644
--- a/inject.js
+++ b/inject.js
@@ -69,10 +69,14 @@ chrome.extension.sendMessage({}, function(response) {
       });

       target.addEventListener('ratechange', function(event) {
-        var speed = this.getSpeed();
-        this.speedIndicator.textContent = speed;
-        tc.settings.speed = speed;
-        chrome.storage.sync.set({'speed': speed});
+        if (event.target.readyState > 0) {
+          var speed = this.getSpeed();
+          this.speedIndicator.textContent = speed;
+          tc.settings.speed = speed;
+          chrome.storage.sync.set({'speed': speed}, function() {
+            console.log('Speed setting saved: ' + speed);
+          });
+        }
       }.bind(this));

       target.playbackRate = tc.settings.speed;
```

However, I want to poke around on other sites with this first to see if it breaks anything.. If you're willing to take it for a spin, that'd be great too. :)

arikroc:
Awesome! Definitely :) 

How would I do that? You make a new branch or upload a modified release, then I uninstall the current version of the extension and manually install the new one?

igrigorik:
[...]

PR #248, opened 2017-07-27
able to change speed before video loads

fent:
As discussed in #233, the fix for that issue makes it so you're unable to set the speed before the video loads. This PR aims to remedy that.

Took the diff from [this comment](https://github.com/igrigorik/videospeed/issues/233#issuecomment-317619644) and added a bit more.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -326,7 +326,7 @@ chrome.extension.sendMessage({}, function(response) {
         } else if (action === 'faster') {
           // Maximum playback speed in Chrome is set to 16:
           // https://cs.chromium.org/chromium/src/media/blink/webmediaplayer_impl.cc?l=103
-          var currentRate = event.target.readyState > 0 ? v.playbackRate : parseFloat(tc.settings.speed);
+          var currentRate = v.readyState > 0 ? v.playbackRate : parseFloat(tc.settings.speed);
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -326,16 +328,20 @@ chrome.extension.sendMessage({}, function(response) {
         } else if (action === 'advance') {
           v.currentTime += tc.settings.advanceTime;
         } else if (action === 'faster') {
+          userAction = true;
           // Maximum playback speed in Chrome is set to 16:
           // https://cs.chromium.org/chromium/src/media/blink/webmediaplayer_impl.cc?l=103
-          var s = Math.min( (v.playbackRate < 0.1 ? 0.0 : v.playbackRate) + tc.settings.speedStep, 16);
+          var currentRate = event.target.readyState > 0 ? v.playbackRate : tc.settings.speed;
```
igrigorik: Where are you grabbing event reference from? =/

fent:
> Have you verified that this solves the problem you described earlier?

Yeah

>We should think how to refactor this.. Branching on boolean flags from different functions is a recipe for subtle bugs down the road.

Do you mean `userAction`? Any ideas how to refactor?

fent:
I've updated this so that it saves the speed when set directly through `runAction()`, since we know those must have been user initiated. Rather than using the `ratechange` event.

I've tested this on Plex web.

igrigorik:
This still requires a major refactoring, closing. If someone wants to pick this back up, we can take another run at it.

PR #251, opened 2017-08-07
Fixed error when using Chrome Autofill. Closes #250

### LINKED ISSUE #250: Error when using browser's form autofill
```
Uncaught TypeError: event.getModifierState is not a function
```

When using native autofill feature in Chrome, `keydown` event is raised but event object is missing `keyCode` property and `getModifierState` function.

What do you say about additional check for these parameters [on the top of event listener](https://github.com/igrigorik/videospeed/blob/master/inject.js#L222)?

p.s. Here you can test the issue: https://www.roboform.com/filling-test-all-fields

p.p.s My Chrome version is `Version 59.0.3071.115 (Official Build) (64-bit)`

vovayatsyuk:
This commit fixes the following error when using form autofill:

```
Uncaught TypeError: event.getModifierState is not a function
```

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -222,7 +222,8 @@ chrome.extension.sendMessage({}, function(response) {
         var keyCode = event.keyCode;
 
         // Ignore if following modifier is active.
-        if (event.getModifierState("Alt")
+        if (!event.getModifierState
+            || event.getModifierState("Alt")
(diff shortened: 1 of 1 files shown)


igrigorik:
Awesome work, thank you!

Issue #199, opened 2017-04-09
Videos don't start on HBO GO when extension is enabled


mariusi7:
I'm trying to see some videos (probably Silverlight?), they work in Incognito when VSC is disabled, but not in normal mode.Trying to see if I can speed up the video on this site but somehow, the extension makes them not play at all.
Example:
https://www.hbogo.ro/content/togetherness-1-family-day-834924115

igrigorik:
> I'm trying to see some videos (probably Silverlight?), they work in Incognito when VSC is disabled, but not in normal mode.

Sounds like the player and extension are not playing nicely together. Unfortunately I can't access the link to test.. are there any public (no login required) videos on that site that we can test?

mariusi7:
There's a section with free vids called "Fara abonament" in the menu, direct link: https://www.hbogo.ro/group/free
The first video was one of them. Try to start from there with any video.
If all fail, maybe you have HBO Go in your country and they have a similar free section and same player ?

igrigorik:
- hbogo.ro gives me an "out of service" error. Guessing, due to geo restrictions.
- http://www.hbo.com/watch-free-episodes - loads + plays fine. Can you confirm?
- http://play.hbogo.com/episode/urn:hbo:episode:GVU47sw4_D1FvjSoJAeYA - clicking "preview" asks me to install Flash -- fail.

Hmm..

mariusi7:
Yes, I confirm the preview from hbo.com loads and can be speed up. The hbogo.com link gives me Access Denied or geo-restricted for direct hbogo.com.
Indeed HBO GO from outside the region give an "out of service", I just tried for hbogo.bg.
So it seems there's no way to you to test this unfortunately :(

igrigorik:
How comfortable are you with Chrome DevTools? Can you inspect the page on your end and check if its a video element or flash? Any exceptions being triggered in the console?

mariusi7:
Do these help you? There are a bunch of errors when it's not playing, some of which repeat when it is (with the extension off). The video is not present in Inspect when it fails to play.
![inspect_when_not_working](https://cloud.githubusercontent.com/assets/27129454/24839495/32efc724-1d64-11e7-9151-127ec35f5e62.png)
![inspect_when_working](https://cloud.githubusercontent.com/assets/27129454/24839493/32ea49a2-1d64-11e7-8f7b-3a15364035da.png)
[...]

PR #252, opened 2017-08-07
Update inject.js

mariusi7:
https://github.com/igrigorik/videospeed/issues/199

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -151,7 +151,7 @@ chrome.extension.sendMessage({}, function(response) {
 
       switch (true) {
         case (location.hostname == 'www.amazon.com'):
-		    case (/www\.hbogo\./).test(location.hostname):
+        case (/www\.hbogo\./).test(location.hostname):
           // insert before parent to bypass overlay
           this.parent.parentElement.insertBefore(fragment, this.parent);
           break;
--- a/inject.js
+++ b/inject.js
@@ -149,14 +149,19 @@ chrome.extension.sendMessage({}, function(response) {
       this.video.classList.add('vsc-initialized');
       this.video.dataset['vscid'] = this.id;
 
-      if (location.hostname == 'www.facebook.com') {
-         // set stacking context to same as parent's parent.
-          // + default fallthrough
-          this.parent.style.zIndex = 'auto';
-      } else if (location.hostname == 'www.amazon.com' || location.hostname.match('www\.hbogo\.*') {
+      switch (true) {
+        case (location.hostname == 'www.amazon.com'):
+		    case (/www\.hbogo\./).test(location.hostname):
           // insert before parent to bypass overlay
           this.parent.parentElement.insertBefore(fragment, this.parent);
-      } else {
+          break;
+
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -149,18 +149,14 @@ chrome.extension.sendMessage({}, function(response) {
...
-
-        case 'www.facebook.com':
-          // set stacking context to same as parent's parent.
+      if (location.hostname == 'www.facebook.com') {
+         // set stacking context to same as parent's parent.
           // + default fallthrough
           this.parent.style.zIndex = 'auto';
-
-        default:
+      } else if (location.hostname == 'www.amazon.com' || location.hostname.match('www\.hbogo\.*') {
```
igrigorik: We should use test() instead of match(), as it returns true or false.

inject.js
```diff
@@ -149,18 +149,14 @@ chrome.extension.sendMessage({}, function(response) {
       this.video.classList.add('vsc-initialized');
       this.video.dataset['vscid'] = this.id;
 
-      switch (location.hostname) {
```
igrigorik: Technically, I think we can do regex matching in switch. Any particular reason why you prefer if's?

mariusi7:
Didn't know JS can have expressions in switch, that's why I used "if" at first. I changed (needed to switch on 'true'). Also used "test()" instead of "match()". 
Note that I couldn't test hotdeploying the switch in debug mode no matter what I tried (either TypeError or thrown out of DevTools). So I  can only hope the syntax is correct.

igrigorik:
> Note that I couldn't test hotdeploying the switch in debug mode no matter what I tried (either TypeError or thrown out of DevTools). So I can only hope the syntax is correct.

I installed the branch locally and everything seems to be working well. That said, I'm not sure I understand the issue above. What do you mean by "hot deploying"? Did you install the extension as an unpacked extension in Chrome? Or, are you just using the GitHub editor to change the code?

mariusi7:
I only used the GitHub editor to change. Was unsure how to test, so I tried to copy-paste the changed switch while stopped in a breakpoint above it in Chrome.

mariusi7:
I'll load it unpacked to make sure it actually works in my testcase...

mariusi7:
Loaded the zip with my branch, disabled the original extension. All worked fine, video played, I could change the speed. So you can merge the changes from my POV if you agree too.

mariusi7:
Unrelated to HBO not playing at all, I noticed now that the speed was reset a few times to 1.0 after I increase and let it play for ~10s. Or it starts at the preferred fast speed and resets to 1.0. This doesn't happen for local files or other video player. 
I read the other open/closed issues about reset.. Do you have any idea why? Buffering? 
It's not a VSC bug, it also resets on this player with it disabled and speed increased with another addon.

igrigorik:
> I read the other open/closed issues about reset.. Do you have any idea why? Buffering?
> It's not a VSC bug, it also resets on this player with it disabled and speed increased with another addon.

This sounds like broken player logic within HBO player. Not much we can do about this..

Thanks for the patch, merging!

Issue #236, opened 2017-07-05
Controller lifetime should be tied to src on video (no src, no controller)


SeanSilke:
There is doubling of controls on vk.com video player.
I am using  0.4.8 version of plugin.
Here is screenshot.
<img width="539" alt="sy8ooncvsbcairmikjhkgw" src="https://user-images.githubusercontent.com/1077074/27861419-81963f98-6189-11e7-9e5c-26670beb9391.png">
You can reproduce this behavior by playing some videos on vk.com pages, for example https://vk.com/iskin

igrigorik:
Can you please try the latest plugin from master? We just landed a fix
yesterday that should address this.

SeanSilke:
Tried with the latest plugin from master. There is still doubling.

igrigorik:
Well, doh.. :-)

Do you have an example URL, ideally something that doesn't require a login?

SeanSilke:
Here is URL https://vk.com/iskin?w=wall475146_35849%2Fall
the login is not required.

igrigorik:
Ah, "fun" one.. The issue is because there are actually two video elements in the container, one on top of the other, and one of them doesn't have a src / servers as a placeholder for in-video ads. As a result, you see two controllers...

![image](https://user-images.githubusercontent.com/10652/27875025-09b487a4-61b2-11e7-9daf-98101daa70cf.png)

This is something I've run into before, but haven't tackled explicitly.. Arguably, we shouldn't initialize the controller for video elements that don't have a src. That said, getting this right will require some careful implementation (and additional MutationObservers on per video element, probably).

canarslan12:
Afaik, video elements can be loaded/played without src with help of js. There might be video.status etc. I hope it will help

canarslan12:
I think second one is not for adds, it is for seek bar's hover images

PR #256, opened 2017-08-28
vk site fix attemp #236

canarslan12:
not sure about all across vk side, it might break something else

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -53,7 +53,7 @@ chrome.extension.sendMessage({}, function(response) {
       if (target.dataset['vscid']) {
         return;
       }
-      if (target.networkState == 0) {
+      if (location.hostname == 'vk.com' && target.networkState == 0) {
         return;
       }
 
--- a/inject.js
+++ b/inject.js
@@ -53,7 +53,7 @@ chrome.extension.sendMessage({}, function(response) {
       if (target.dataset['vscid']) {
         return;
       }
-      if (location.hostname == 'vk.com' && target.networkState == 0) {
+      if (target.networkState == 0) {
         return;
       }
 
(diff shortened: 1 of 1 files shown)


SeanSilke:
It is quite good.

igrigorik:
What problem is this solving? 

FWIW, I would strongly prefer not to start littering code with origin ifdefs. Is there a more generic solution to a problem you're trying to solve?

SeanSilke:
@igrigorik This is solving https://github.com/igrigorik/videospeed/issues/236.

igrigorik:
Ah, I see.. interesting. Couple of notes..

- Agree that this shouldn't be vk.com specific
- Current logic prevents controller from initializing if networkState is 0 when it's appended to the DOM (which makes sense), but it also does not account for case where networkState changes later. Is there a callback that we can or should listen to to invoke the registration?
  - As this, this would prevent VC from starting on any 'lazy-loaded' videos.
- There's some conceptual overlap here with https://github.com/igrigorik/videospeed/commit/bbb8a7354b6e9ff6e9cf825f5d9dd89c5a87e60c - we should think about if/how we can refactor that with this work.

canarslan12:
I think vk.com using 'lazy-loaded' videos and no problem happening.
do you suggest me to add `target.addEventListener('_networkStatechange(?)_', function(event) {`?

igrigorik:
Yes, except.. I don't believe there is such a callback, which is part of the problem?

canarslan12:
we can use some of them but I'm not sure about it, i think onprogress will be better. 
> loadstart
> durationchange
> loadedmetadata
> loadeddata
> progress
> canplay
> canplaythrough
also we can do this for all sites which use lazyloading differently. It will be good enhancement for extension.

igrigorik:
@canarslan12 can we remove the vk.com clause in this PR? What will it break? As stated earlier, I'd like to avoid site-specific rules whenever possible.

igrigorik:
Closing. If this is still an issue for vk, let's open a new bug and discuss the possible approaches there.

PR #267, opened 2017-09-24
Update inject.js for HBOGO (2)

mariusi7:
Some regional HBO GO sites don't use www anymore, so issue 199 reappeared.


igrigorik:
👍  thanks!

PR #270, opened 2017-09-29
Remove vscid even if ctrl is not found, fixes new Plex web

fent:
This fixes an issue with the new Plex web app in which the speed controller will only show up for the first video, but not any after that.

It looks like this happens because when a video finishes, or is exited, its original parent is removed, which removes the wrapper vsc creates.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -267,9 +267,9 @@ chrome.extension.sendMessage({}, function(response) {
               let id = node.dataset['vscid'];
               let ctrl = document.querySelector(`div[data-vscid="${id}"]`)
               if (ctrl) {
-                node.classList.remove('vsc-initialized');
                 ctrl.remove();
               }
+              node.classList.remove('vsc-initialized');
               delete node.dataset['vscid'];
             }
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -268,9 +268,9 @@ chrome.extension.sendMessage({}, function(response) {
               let ctrl = document.querySelector(`div[data-vscid="${id}"]`)
               if (ctrl) {
                 node.classList.remove('vsc-initialized');
```
igrigorik: It's odd that we'd only remove data attr outside of ctrl check. Should the classList remove be outside of the ctrl check as well? Would that break anything?

fent: It wouldn't. The class is added at the spot that the data attr is. However, only the attr is checked to see if a video has already been initialized.

Not sure what the class is used for. Unless it's used to style the video in some way, but I can't find the css for that either.

igrigorik: For consistency, let's move both to same place -- otherwise, lgtm.

PR #292, opened 2017-11-12
better way of following mouse on controller drag

fent:
Hi,

This PR fixes a minor annoyance when dragging the video speed controller.

1) On some sites like Plex and Netflix, they have placed elements over the video. Sometimes when dragging the controller too fast, the mouse can go over these elements, and the dragging will end. After trying a few things, I settled on the following solution.

When dragging begins, it keeps traversing up the DOM until it finds a parent with the same dimensions as the initial parent of the video. With this method, the mouse has a higher chance of not accidentally being dragged out of the containing element, since the containing element will now be higher in the DOM tree.

This method works with Plex, Netflix, and Facebook. If there are other common sites I should test, please let me know.

2) When dragging the controller in a fast manner, over time, the mouse position can get unsynchronized with the controller. For this, the initial mouse position, along with the initial controller position, are used to determine where the controller will be placed on every `mousemove`. Rather than relying on `e.movementX` and `e.movementY`.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -133,12 +133,12 @@ chrome.runtime.sendMessage({}, function(response) {
       `;
       shadow.innerHTML = shadowTemplate;
       shadow.querySelector('.draggable').addEventListener('mousedown', (e) => {
-        runAction(e.target.dataset['action'], document);
+        runAction(e.target.dataset['action'], document, false, e);
       });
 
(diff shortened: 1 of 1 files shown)


igrigorik:
I've been running this locally for past week, everything seems to be working as advertised. Great stuff!

PR #305, opened 2018-01-03
document Shift+S workaround for Vimeo

### LINKED ISSUE #304: Shift+S workaround for Vimeo needs documenting
Hi there, I'm new to this great extension, and swiftly discovered the clash described in #69.  However I couldn't see any workaround documented, so I had to go digging in the GitHub issues to find it.  I'll submit a PR to fix that.

aspiers:
Fixes #304.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ Once the extension is installed simply navigate to any page that offers HTML5 vi
 * **X** - advance video by 10 seconds.
 * **V** - show/hide the controller.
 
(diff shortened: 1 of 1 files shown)


igrigorik:
Looks good. Although, given that we also have other shortcuts (e.g. X) that are affected, perhaps rewrite this to be more generic ~~ if shortcut clashes, use Shift-Key instead. This applies outside of Vimeo as well.

Issue #315, opened 2018-02-21
Unable to click/move controls on v.redd.it videos


AUTplayed:
If you view any video on https://www.reddit.com/domain/v.redd.it/  you can see that the controls are 

- in the middle of the video
- unclickable

which is very annoying. I would love for the controller to be either positioned at the top left corner like in other videos or click/moveable.

AUTplayed:
Is this not live yet or did reddit change something? Anyways it doesn't work for me, still in the middle of the screen and not moveable/clickable

PR #325, opened 2018-04-01
unclickable overlay fix on reddit video - #315

canarslan12:
not fixing "in the middle of the video" part

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -95,8 +95,8 @@ chrome.runtime.sendMessage({}, function(response) {
     tc.videoController.prototype.initializeControls = function() {
       var document = this.document;
       var speed = parseFloat(tc.settings.speed).toFixed(2),
-        top = "0px",
-        left = "0px";
+        top = Math.max(this.video.offsetTop, 0) + "px",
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -95,8 +95,8 @@ chrome.runtime.sendMessage({}, function(response) {
     tc.videoController.prototype.initializeControls = function() {
       var document = this.document;
       var speed = parseFloat(tc.settings.speed).toFixed(2),
-        top = Math.max(this.video.offsetTop, 0) + "px",
-        left = Math.max(this.video.offsetLeft, 0) + "px";
```
igrigorik: Hmm, why did we need to drop the offset's? 

Is the fix here the special case in the switch below, or the code above?

canarslan12: They are causing location bug and it's useless if I'm correct

canarslan12: switch below: fix for unclickable overlay
code above(offset): fix for location of overlay

igrigorik:
Thanks for the nudge.

Issue #326, opened 2018-04-02
Controls no longer show or work on local videos (and some other sites) v0.5.1


mariusi7:
I guess there was an update since tomorow, but the controls don't show anymore on local videos for me. Settings are ok, "Allow access to file URLs" is ON, nothing changed from my part since yesterday when it still worked. Other users complain about certain websites (on the Chrome page).

mariusi7:
Chrome version: 65.0.3325.181
 (unchanged since last week)
Oddly, I tried installing 1.5.0 in dev mode but it still doesn't work (?). I'll keep trying

mariusi7:
Version 0.49, id cjfjblcfekkjldlnkccaafonahammakg, commit 194fcc107680f8d84ca75749c375c152f991eedd works.

mariusi7:
@igrigorik Please have a look. Thx!

mariusi7:
@canarslan12 So does it work for you (in latest Chrome version) for local files or not ?
It's not just me, I saw a couple of other users mentioning problems in the last 6h ( https://chrome.google.com/webstore/detail/video-speed-controller/nffaoalbilbmmfgbnbgppjihopabppdk/support?hl=en )

canarslan12:
can you try this fixed version?

mariusi7:
I confirm that your version with commit dd7d976e90af60e378aa68b712448837594b6193 works again for my local files in latest Chrome version.

apurplecaboose:
the extension still does not work for me on local files for some reason...

mariusi7:
@apurplecaboose did you install the unpacked extension from a downloaded file videospeed-dd7d976e90af60e378aa68b712448837594b6193.zip ? It shows at 0.5.0 (ID: foacgnafkoecldehmhofegjcldbkgjph)
You can also check by opening Developer Tools, tab Sources, subtab Content Scripts and line 194 needs to look like here: https://i.imgur.com/lERmyfr.png

alon91:
Could you help with guidance on steps that we need to follow to download the fixed version? Haven't been able to figure it out on my own. Many thanks in advance!

mariusi7:
[...]

Issue #309, opened 2018-02-01
Cannot read property 'body' of null


wesbos:
Getting this error on Chrome Canary. Any idea why it would happen?

I can get it on wesbos.com and localhost as well as many other sites - but not every site. 

![](http://wes.io/pBoA/content)

wesbos:
I can also confirm it happens in stable Chrome

igrigorik:
Hey Wes. It looks like your site is powered by WordPress.. Are you using Videopress to play the videos? If so, there is an outstanding issue here: https://github.com/igrigorik/videospeed/issues/194

wesbos:
Hey! Thanks for the reply. Both sites don’t have videos on them. Only one of them is Wordpress as well. Do you see the error when visiting my site?

igrigorik:
Ah, doh, yes.. sorry misdiagnosed this. Looks like we have a null document getting passed into:

https://github.com/igrigorik/videospeed/blob/master/inject.js#L205-L210

wesbos:
Very bizarre - do you think I could just return from that function if there is no document? It's working totally fine on all pages with videos

igrigorik:
I believe this should be fixed in master. If you're willing, try the unpacked extension? :)

wesbos:
That fixed it! Thank you :D

berekuk:
@igrigorik, do you plan to release this fix to chrome web store any time soon? It doesn't break anything, but it's quite annoying when I'm trying to debug my own code.

igrigorik:
@berekuk I'll push a new version live this week — thanks for the nudge. Need to triage a few new issues, but hopefully that won't take long.

wesbos:
please do - I've had to keep disabling the extension when I'm working on my own code

canarslan12:
I cannot reproduce this one @igrigorik

PR #327, opened 2018-04-02
Fix for #326

canarslan12:
I hope that you didn't change this for some reason
https://github.com/igrigorik/videospeed/commit/e50c29c5dd521cb0b91ffcc522bfa3e54e002070

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -191,7 +191,7 @@ chrome.runtime.sendMessage({}, function(response) {
     window.onload = () => {
       initializeNow(window.document)
     };
-    if (document && document.doctype && document.doctype.name == "html") {
+    if (document) {
       if (document.readyState === "complete") {
         initializeNow(document);
       } else {
--- a/inject.js
+++ b/inject.js
@@ -206,7 +206,7 @@ chrome.runtime.sendMessage({}, function(response) {
 
   function initializeNow(document) {
       // enforce init-once due to redundant callers
-      if (document.body.classList.contains('vsc-initialized')) {
+      if (!document.body || document.body.classList.contains('vsc-initialized')) {
         return;
       }
       document.body.classList.add('vsc-initialized');


inject.js
```diff
@@ -191,7 +191,7 @@ chrome.runtime.sendMessage({}, function(response) {
     window.onload = () => {
       initializeNow(window.document)
     };
-    if (document && document.doctype && document.doctype.name == "html") {
```
igrigorik: Why is this being removed? This was added to address https://github.com/igrigorik/videospeed/issues/309

mariusi7: I think that test makes the initialization not happen when the URL is smth like file:///C:/Users/myvideo.mp4 (?)

mariusi7: https://i.imgur.com/DeJZjiN.png

igrigorik: Ah, interesting.. doh. Yeah, let's find a different solution for #309 then. Could you take a look at the page flagged in #309 and see if we can find a different solution for it?

canarslan12: I think my solution is not breaking it but might break it. why is it here?
https://github.com/igrigorik/videospeed/commit/7606224ec92790f65f27b348c9fb7346ecbc7d3e

canarslan12:
I think it's good now

mariusi7:
@igrigorik Please check and approve this merge. 
When can we expect a new final version that contains the fix for #326 ?

igrigorik:
Sorry about the delay. Testing on this end.. seems to be working as promised!

mariusi7:
Any plans for a new version 0.5.2 with this fix contained ? I'm using the development mode which always nags me on open

igrigorik:
@mariusi7 yes, planning to release an update soon. Need to track down a few (unrelated, I believe) issues first.

PR #334, opened 2018-05-12
fixes #332 with adding keydown listener to parent

### LINKED ISSUE #332: Doesn't work for new Twitch Clips layout
Twitch Clips has a new layout as of a few days ago. [Example here](https://clips.twitch.tv/ExcitedWealthyGazelleKippa).

VSC has stopped working for it now. It will work if you click on the video though.

The old Twitch Clip layout just included the video tag in the page's HTML I think, so nothing special there. But now, they embed the video via an `iframe`, then dynamically load the video. VSC should work for iframes, but since the video here is loaded dynamically, I guess it doesn't catch it on the initial page load.

canarslan12:
fixes #332
this adds keydown listener to parent document if there is one.
it does not use background files which uses unnecessary ram

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -203,7 +203,13 @@ chrome.runtime.sendMessage({}, function(response) {
       }
     }
   }
-
+  inIframe () {
+    try {
+      return window.self !== window.top;
+    } catch (e) {
+      return true;
+    }
+  }
   function initializeNow(document) {
       // enforce init-once due to redundant callers
       if (!document.body || document.body.classList.contains('vsc-initialized')) {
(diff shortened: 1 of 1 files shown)


garyking:
@igrigorik It works 👍

igrigorik:
Yay! Thanks guys.

garyking:
Has this change actually been deployed? The Chrome store shows that this extension was updated April 1, yet I don't  have the problem on Twitch Clips anymore. Yet I still have the problem that #325 supposedly fixes.

canarslan12:
it is half fixed(just clickable) after April 1st and not deployed now.
________________________________
Gönderen: garyking <notifications@github.com>
Gönderildi: 10 Haziran 2018 Pazar 00:03
Kime: igrigorik/videospeed
Bilgi: Can Arslan; Author
Konu: Re: [igrigorik/videospeed] fixes #332 with adding keydown listener to parent (#334)


Has this change actually been deployed? The Chrome store shows that this extension was updated April 1, yet I don't have the problem on Twitch Clips anymore. Yet I still have the problem that #325<https://github.com/igrigorik/videospeed/pull/325> supposedly fixes.

Issue #345, opened 2018-06-16
Controller not working on embeds or local videos


igrigorik:
Example embeds:

- https://www.igvita.com/2012/09/12/web-fonts-performance-making-pretty-fast/
- Ditto for vimeo embeds

@canarslan12 I suspect this is due to one of the recent changes we've landed, but haven't tracked it down yet.

canarslan12:
it is happening because it wants to addEventListener to parent, which is cross origin. it is not allowed due to cross origin policy. I'm just pull requesting something which fix problem with try catch block but it is not working if I push keys in parent document (it was like that in previous versions)

also, you look nice on the video 👓

igrigorik:
Hmm, testing with #346: same issue as before. The controller is not coming up.

igrigorik:
On second look, just realized that this a dupe of https://github.com/igrigorik/videospeed/issues/326 — let's continue there.

PR #346, opened 2018-06-16
Controller not working on embeds #345

canarslan12:
#345


igrigorik:
Thanks for the quick quick. As a meta request / point: please use more descriptive commit messages, and we should also consider adding code comments as well. This way we can more easily page back why certain logic exists in these places.

Issue #347, opened 2018-06-16
Allow user to define custom shortcuts


igrigorik:
Currently we provide a predefined list of shortcuts, with each additional shortcut requiring an explicit code update as well as definition of a default key. This does not scale and causes shortcut conflicts. Instead of using a frozen list of actions, against which you can assign a hotkey..

Refactor / implement a mechanism where the user can add own shortcuts in the options panel:

- Click "add shortcut" presents:
  - Input box for the shortcut key
  - Dropdown of allowed actions — e.g. play, pause, seek, ratechange
    - Selecting some actions would inject another "value" box — e.g. value for ratechange.
- There should be a "x" / remove action option to remove assignments.

We can still ship a default set of shortcuts, matching what we have today, but this allows the user to then define as many shortcuts as they want with as many value variations as they want.

canarslan12:
https://github.com/canarslan12/videospeed/tree/testing

I started to work on it and created tc.settings.keyBindings array which contains `whattodo(rewind etc.)-keyCode-value-force` as an arrays for each key bindings.
Code is unexplained and not-cleaned, I will keep working probably next week. If someone has any suggestions, I can add it before finishing it. 
force field is for trying to stop default function of button such as space pausing in twitch.

igrigorik:
Resolved via #350.

PR #350, opened 2018-06-27
Allow user to define custom shortcuts

canarslan12:
- Over-all working structure changed
- tc.settings.keyBindings array added (whattodo(rewind etc.)-keyCode-value-force)
- Options page changed (Jquery Added)
New settings page:
image
#347


canarslan12:
bug:
it does not successfully ignore in text areas

options.html
```diff
@@ -3,67 +3,90 @@
 	<head>
 		<title>Video Speed Controller: Options</title>
 		<link rel="stylesheet" href="options.css" />
-		<script src="options.js"></script>
+        <script src="jquery-3.2.1.slim.min.js"></script>
```
igrigorik: Could we avoid it? :-)

options.js
```diff
@@ -1,32 +1,35 @@
...
-  fasterKeyCode: 68,    // default: D
-  fastKeyCode: 71,      // default: G
-  rewindKeyCode: 90,    // default: Z
-  advanceKeyCode: 88,   // default: X
+  speed: 1.0,           // default:
   displayKeyCode: 86,   // default: V
   rememberSpeed: false, // default: false
   startHidden: false,   // default: false
+  keyBindings: [
+    ["slower", 83, 0.1, 0], // default: S 0.1
```
igrigorik: To improve readability / ergonomics, let's use an object instead? 

`
[ 
  { "action": "slower", "keycode": ...}, 
  ...
] 
`

options.js
```diff
@@ -1,32 +1,35 @@
...
+    ["faster", 68, 0.1, 0], // default: D 0.1
+    ["rewind", 90, 10, 0], // default: Z 0.1
+    ["advance", 88, 10, 0], // default: X 0.1
+    ["reset", 82, 1, 0], // default: R 1
+    ["fast", 71, 1.8, 0] // default: G 1.8
+  ],
+  version: "0.5.3",     // default: 0.5.3
```
igrigorik: why do we need to mirror the version number here?

options.js
```diff
@@ -1,32 +1,35 @@
...
   blacklist: `
     www.instagram.com
     twitter.com
     vine.co
     imgur.com
-  `.replace(/^\s+|\s+$/gm,'')
+  `.replace(/^\s+|\s+$/gm, '')
 };
 
+var keyBindings = []; // whattodo-keyCode-value-force
```
igrigorik: not sure what this comment means? :)

options.js
```diff
@@ -88,88 +94,123 @@ function updateShortcutInputText(inputId, keyCode) {
   document.getElementById(inputId).keyCode = keyCode;
 }
 
+function updateCustomShortcutInputText(inputItem, keyCode) {
+  inputItem.value = keyCodeAliases[keyCode] || String.fromCharCode(keyCode);
+  inputItem.keyCode = keyCode;
+}
+
+function add_shortcut() {
+  var html = '<select class="customDo">' +
```
igrigorik: Let's use template literals here, it'll make it much easier to maintain: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

options.js
```diff
@@ -88,88 +94,123 @@ function updateShortcutInputText(inputId, keyCode) {
...
+    '<option value="slower">Decrease speed</option>' +
+    '<option value="faster">Increase speed</option>' +
+    '<option value="rewind">Rewind</option>' +
+    '<option value="advance">Advance</option>' +
+    // '<option value="reset">Reset speed</option>' +
```
igrigorik: Any particular reason why this one is disabled?

options.js
```diff
@@ -88,88 +94,123 @@ function updateShortcutInputText(inputId, keyCode) {
...
+    '<button class="removeParent">X</button>';
+  var div = document.createElement('div');
+  div.setAttribute('class', 'row customs');
+  div.innerHTML = html;
+  var customs_element = document.getElementById("customs");
+  customs_element.insertBefore(div, customs_element.children[customs_element.childElementCount - 1]);
+}
+
+function createKeyBindings() {
+  var doo = $(this).find(".customDo").val();
```
igrigorik: We can use querySelector here, no need for jQuery.

See: https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

options.js
```diff
@@ -88,88 +94,123 @@ function updateShortcutInputText(inputId, keyCode) {
...
-  advanceKeyCode = isNaN(advanceKeyCode) ? tcDefaults.advanceKeyCode : advanceKeyCode;
-  slowerKeyCode = isNaN(slowerKeyCode) ? tcDefaults.slowerKeyCode : slowerKeyCode;
-  fasterKeyCode = isNaN(fasterKeyCode) ? tcDefaults.fasterKeyCode : fasterKeyCode;
-  fastKeyCode   = isNaN(fastKeyCode) ? tcDefaults.fastKeyCode : fastKeyCode;
+  var blacklist = document.getElementById('blacklist').value;
+  var version = tcDefaults.version;
+
   displayKeyCode = isNaN(displayKeyCode) ? tcDefaults.displayKeyCode : displayKeyCode;
 
+  chrome.storage.sync.clear();
```
igrigorik: Why do we need the clear()?

options.js
```diff
@@ -88,88 +94,123 @@ function updateShortcutInputText(inputId, keyCode) {
...
-    fastKeyCode:    fastKeyCode,
     displayKeyCode: displayKeyCode,
-    rememberSpeed:  rememberSpeed,
-    startHidden:    startHidden,
-    blacklist:      blacklist.replace(/^\s+|\s+$/gm,'')
-  }, function() {
+    rememberSpeed: rememberSpeed,
+    startHidden: startHidden,
+    keyBindings: keyBindings,
+    version: version,
```
igrigorik: What's the utility of storing version?

options.js
```diff
@@ -88,88 +94,123 @@ function updateShortcutInputText(inputId, keyCode) {
...
-    updateShortcutInputText('fastKeyInput', storage.fastKeyCode);
+  chrome.storage.sync.get(tcDefaults, function (storage) {
     updateShortcutInputText('displayKeyInput', storage.displayKeyCode);
     document.getElementById('rememberSpeed').checked = storage.rememberSpeed;
     document.getElementById('startHidden').checked = storage.startHidden;
     document.getElementById('blacklist').value = storage.blacklist;
+
+    for (let i in storage.keyBindings) {
+      var item = storage.keyBindings[i];
+      if (i < 6) {
```
igrigorik: Instead of relying on indexes (this will break and cause subtle bugs), let's annotate the keybindings with "default" or "visible" flag.

options.html
```diff
@@ -3,67 +3,90 @@
...
-    </section>
+        <div class="row customs">
+            <select class="customDo">
+                <option value="slower">Decrease speed</option>
+            </select>
+            <input class="customKey" type="text" value="" placeholder="press a key">
+            <input class="customValue" type="text" placeholder="value (0.10)">
+            <select class="customReturn">
+                <option value="0">Do not disable website key bindings</option>
+                <option value="1">Disable websites key bindings</option>
```
igrigorik: ^ what are these for? In practice we cannot guarantee that we can disable or override site-provided keybindings.

options.html
canarslan12: for making/testing faster I added jquery. I'm not sure about if I can edit all usages easily. Do we need to remove jquery since it is just for settings page?

options.js
canarslan12: on line 43 `if (!storage.version) // if version is lower than 0.5.2 or first initialization` we need to use something to test if it's first upgrade. maybe we can try if we have keyBindings in storage

options.js
canarslan12: that way we should also check for default flag for buttons (which is not working now)
what do you mean with visible? visible in overlay? It might be good that way

options.html
canarslan12: it is an experimental feature which is good for pause. that way we can use space bar to pause video even in livestreams(like twitch).

options.html
igrigorik: Which are the parts that would be more challenging to replace?

options.html
igrigorik: I'd propose we decouple experimental features from this PR. 

If we ship this, the natural expectation of the user is that we can reliably "disable website key bindings". In my experience, that's neither possible nor necessarily desirable. 

Can we pull this out and start a new thread to discuss what you're after here?

options.js
igrigorik: Hmm, I don't see that line?

If we do want to persist the version, we can use `chrome.runtime.getManifest().version` to query the version in the manifest, instead of hardcoding it.

options.js
igrigorik: Can we make this conditional? As in, I do see the utility of having a ~migrate function between versions, but I don't think we need to clear this on every update? 

E.g. if previously persisted version != current version, run clear before persisting new set.

options.js
igrigorik: Given that we provide this option today, I think we should keep. I'm sure folks will complain otherwise :)

options.js
igrigorik: 'visible' might not be the best name. I just mean that we shouldn't rely on position indexes and instead add an explicit flag for any shortcuts that have special treatment. We can use 'required', or 'default', or ...

options.html
canarslan12: I tried it and it has no problem like that. I can not take this out of this pr since I developed everything together

options.js
```diff
@@ -185,18 +226,26 @@ document.addEventListener('DOMContentLoaded', function () {
...
-  initShortcutInput('slowerKeyInput');
-  initShortcutInput('fasterKeyInput');
-  initShortcutInput('fastKeyInput');
   initShortcutInput('displayKeyInput');
 
-  document.getElementById('rewindTime').addEventListener('keypress', inputFilterNumbersOnly);
-  document.getElementById('advanceTime').addEventListener('keypress', inputFilterNumbersOnly);
-  document.getElementById('speedStep').addEventListener('keypress', inputFilterNumbersOnly);
-  document.getElementById('fastSpeed').addEventListener('keypress', inputFilterNumbersOnly);
+  $(document).on("keypress", ".customValue", inputFilterNumbersOnly);
```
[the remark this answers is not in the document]
canarslan12: yes I was looking for it but could make it work. now it works without jquery

options.html
canarslan12: since it's experimental, I hide .customForce but code in back is still there

inject.js
```diff
@@ -1,55 +1,124 @@
...
+      resetKeyCode: null,    // default: R
+      slowerKeyCode: null,    // default: S
+      fasterKeyCode: null,    // default: D
+      rewindKeyCode: null,    // default: Z
+      advanceKeyCode: null,   // default: X
+      fastKeyCode: null,      // default: G
+      /**
+       * these are not used
+       * but should be stay there because chrome.storage.sync.get needs them.
+       */
```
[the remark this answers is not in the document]
canarslan12: It's like opening and closing tags

inject.js
```diff
@@ -1,55 +1,124 @@
...
-      resetKeyCode:  82,    // default: R
-      slowerKeyCode: 83,    // default: S
-      fasterKeyCode: 68,    // default: D
-      rewindKeyCode: 90,    // default: Z
-      advanceKeyCode: 88,   // default: X
+
+      /**
+       * these are not used
+       * but should be stay there because chrome.storage.sync.get needs them.
+       */
```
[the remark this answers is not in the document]
canarslan12: Yes

inject.js
```diff
@@ -204,158 +274,159 @@ chrome.runtime.sendMessage({}, function(response) {
...
-            runAction('display', document, true)
-          } else if (keyCode == tc.settings.fastKeyCode) {
-            runAction('fast', document, true);
+        if (keyCode == tc.settings.displayKeyCode) {
+          runAction('display', document, true)
+        }
+        var item = tc.settings.keyBindings.find(item => item.key === keyCode);
+        if (item) {
+          runAction(item.action, document, item.value);
+          if (item.force === "true") {// disable websites key bindings
```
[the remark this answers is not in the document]
canarslan12: On my tests it's always first one but it's disabled in settings page for now. Because it also stop everything like space to page down even though there is no video

inject.js
```diff
@@ -365,41 +436,69 @@ chrome.runtime.sendMessage({}, function(response) {
...
-      chrome.storage.sync.set({'resetSpeed': v.playbackRate});
+      setKeyBindings("reset", v.playbackRate);// resetSpeed
+      // chrome.storage.sync.set({'resetSpeed': v.playbackRate});
       v.playbackRate = target;
     }
   }
 
+  function muted(v, value) {
+    v.muted = v.muted !== true; //reverse muted status
+    /* this can be used if someone wants just mute button
```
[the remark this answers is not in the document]
canarslan12: What do you mean by rest? Value parameter?

canarslan12:
I think it's ready to merge

PR #392, opened 2018-11-27
fixes #381: YouTube full screen title overlap

### LINKED ISSUE #381: partially incompatible with YouTube: in fullscreen mode, video title partially overlaps videospeed controls
steps to reproduce the problem:
1. open any youtube video on youtube website in desktop browser (e. g. google chrome)
2. switch to video fullscreen mode (F from keyboard, not F11)
3. place your mouse pointer on videospeed controls, especially move it on top part of the controls panel

what is the expected result:
it is displayed as usual

actual result:
it gets hidden because mouse pointer touches video title

motivation:
currently, youtube is world's most popular video hosting service, therefore, imo, the plugin should be fully compatible with it, shouldn't it?

fix proposal:
imo, move the default controls panel positon a couple of pixels down would easily fix the problem. i. e. currently, its default position is approximately 10px under the top of the video, i think 20px would be enough.

video screencast (gif) attached below
ezgif com-video-to-gif
at the screencast you can see the controls are at their initial position first, the panel disappears when cursor gets close to video title, then i move the panel a little bit lower and it doesn't disappear anymore

canarslan12:


inject.css
```diff
@@ -26,7 +26,7 @@
 /* e.g. https://www.igvita.com/2012/09/12/web-fonts-performance-making-pretty-fast/ */
 .html5-video-player:not(.ytp-hide-info-bar) .vsc-controller {
   position: relative;
-  top: 40px;
```
igrigorik: Is there a way to target this to fullscreen mode only? Otherwise this offsets the controller on all embeds, which looks a bit wonky.

canarslan12: ytp-hide-info-bar is only visible in normal screen on youtube.com
since they changed embed's style, we can also target embed with or without fullscreen. (channel logo on title bar)

Issue #370, opened 2018-08-27
Youtube speed resets every time while toggling Full Screen and back.


BerkayTamer:
Apparently, this speed-reset issue has been happening when toggling the theater mode but now this issue has recently started to happen when opening the Full Screen, as well as going back to the normal screen from the Full Screen. I go on Full Screen and back often which is why this has created some annoyance for me and I am sure I am not the only one getting annoyed by this.

Also, hopefully, this info will be helpful when it comes to fixing this issue:

 I believe this issue has started to happen after youtube added this **"Scroll for details"** function on Full-Screen videos where you get to scroll down and see the comments/related videos even on Full-Screen mode.

<img width="189" alt="scroll for details" src="https://user-images.githubusercontent.com/17339988/44675143-a5e70180-a9fd-11e8-9862-b2347a583a4e.png">

I guess I can get used to this but it would've been amazing to see this problem get fixed, hopefully, it has an easy fix. Btw, I am using Chrome on a Macbook Pro and didn't test the issue on other browsers/systems but I assume they would have the same outcome due to Youtube's recent change on Full-Screen mode. I should also say that **Video Speed Controller** is my most favorite browser extension ever and I am so grateful to its creator, saved me so much time!

igrigorik:
Hmm, can't see to reproduce this on this end. Do you have an example URL that you can share?

drewta:
Have the same issue with speed resetting to 1.0 when toggling between normal and theater mode on any video (e.g. https://www.youtube.com/watch?v=0pO7DWIfR6Y) using either hotkey (T) or manually clicking the button. Doesn't happen when toggling between normal and full screen. I also don't have the scroll for details function, so unsure if that's related. Using Chrome 68.0.3440.106 on a Windows 10 device.

BerkayTamer:
Oh, I just realized it works on everywhere except when I open Youtube on incognito mode and while being signed into my Youtube account. The weird thing is that it does work when I am not signed into my account on incognito mode. 

So in every video that I tested, for example this one: https://www.youtube.com/watch?v=LnQukViKyGY 

When I am in incognito mode and signed into my Youtube account, the speed that I set resets while opening the video on Full Screen or going back from Full Screen to normal screen again. However, since this issue only happens on incognito mode while being signed into a Youtube account, I guess this is not a big issue overall. I mean it's a big issue for me since I use one account on normal Chrome while using my other youtube account on the incognito mode but I guess I can get used to this.

[...]

PR #400, opened 2018-12-09
YouTube toggling Full Screen bug fix

canarslan12:
Fix for #370

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -277,19 +277,36 @@ chrome.runtime.sendMessage({}, function(response) {
         }, true);
       });
 
+      var timer = null, last_vscid = null;
+
+      function deleteCtrl(node) {
+        let id = node.dataset['vscid'];
+        let ctrl = document.querySelector(`div[data-vscid="${id}"]`)
+        if (ctrl) {
+          ctrl.remove();
+        }
+        node.classList.remove('vsc-initialized');
+        delete node.dataset['vscid'];
+      }
+
       function checkForVideo(node, parent, added) {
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -277,19 +277,36 @@ chrome.runtime.sendMessage({}, function(response) {
...
-              let id = node.dataset['vscid'];
-              let ctrl = document.querySelector(`div[data-vscid="${id}"]`)
-              if (ctrl) {
-                ctrl.remove();
+              if (timer_new) {
+                // already waiting other one, delete this one immediately
+                deleteCtrl(node)
+              } else {
+                // wait for new one
+                timer_new = setTimeout(deleteCtrl, 10, node)
```
igrigorik: Hmm, not sure I follow how this works.. There is no guarantee that video element being removed is the same element that's being added later to the DOM, so how does setting the timeout here help address the problem? 

I feel like I'm missing something obvious :)

canarslan12:
Even though its not garantee, we can just wait for 10 ms if new one created. If there's no new one it will destroy it as usual. If there is new one, it clear timeout and do nothing (which is working strangely)

igrigorik:
Is that because YT is recycling the same video element? If so, tackling https://github.com/igrigorik/videospeed/issues/236 would be the appropriate fix for it.

igrigorik:
I suspect the player is removing the node from DOM and then immediately reinserting it, which is why this logic works.. because the controller is not destroyed and tied to same vscid instance.

Issue #368, opened 2018-08-04
Plex: Only works on first played video.


Tenerezza:
It works pretty well overall with Plex web player ( https://www.plex.tv/ ), the issue is that it only works for the first video you play, the second+ video the control does not show up nor does the shortcut works, it also revert back to the initial speed

bbobrian:
I'm seeing similar behavior, which I've seen previously and then it seemed to resolve itself. But now it's back. Not sure if it's due to updates in PlexWeb or Chrome. However, I have noticed that the second+ video does play at the same speed as the previous video, even though the videospeed controls do not appear.

sflesch:
Glad to know it's not just me. Hoping it gets looked at. It's been bugging me for a while, but I finally decided to see what was going on.

Pythor:
Still experiencing this, and it kills me...  If we can't stop Plex from unloading the extension, is there any way to add a hotkey to reload the extension without having to reload the whole page?

captainevdo:
Unfortunately, I think this is due to either plexweb or chrome as bbobrian mentioned.  When this issue arose last year, the dev released a quickfix prior to wide release.  By running the developer extension I was able to circumvent the issue.  Since it rearose, I went back to that unpacked version and it persists, so it isn't do to changes made by igrigorik.  I've posted in the plexforums to see if anyone over there has solved it.  Pythor's idea of a quick extension reload is a good one though.

sflesch:
Did you find anything on Plex? I looked a bit, but didn't get far.

captainevdo:
> Did you find anything on Plex? I looked a bit, but didn't get far.

Made a new post because I didn't find anything germane to the topic.  Will keep everyone updated if there are responses.

captainevdo:
The problem has now extended to the main server player, not just plexweb.... =/

sflesch:
What's the difference between server player and plexweb?

captainevdo:
I was being a bit opaque with my terms there, sorry about that.  One being the actual user interface of my plex media server that I open locally on my network via IP etc, the other being the player I get remotely at <plex.tv/web/app>.  Up until a day or so ago, it was working on my server at home even though the remote one wasn't; now borked on both.

Hendo-AU:
Have had the same problem for a few weeks now since they updated the Plex UI

sflesch:
@captainevdo Gotcha. I usually watch on the same PC I serve from. 2mbps isn't great for serving outside.

[...]

PR #403, opened 2018-12-14
Plex Tv Fix

canarslan12:
Fix for Plex: Only works on first played video. #368
confirmation and futher tests needed.
some sites to test:
- [x] youtube
- [ ] facebook https://www.facebook.com/facebook/videos/568560330227157/
- [ ] twitter https://twitter.com/igrigorik/status/747826527295811584
- [ ] reddit https://www.reddit.com/domain/v.redd.it/
- [ ] vk https://vk.com/iskin?w=wall475146_35849%2Fall

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -119,7 +119,7 @@ chrome.runtime.sendMessage({}, function(response) {
 
   function defineVideoController() {
     tc.videoController = function(target, parent) {
-      if (target.dataset['vscid']) {
+      if (target.dataset['vscid'] && document.querySelector(".vsc-controller")) {
         return;
       }
 


inject.js
```diff
@@ -119,7 +119,7 @@ chrome.runtime.sendMessage({}, function(response) {
 
   function defineVideoController() {
     tc.videoController = function(target, parent) {
-      if (target.dataset['vscid']) {
+      if (target.dataset['vscid'] && document.querySelector(".vsc-controller")) {
```
igrigorik: What's the logic here? Or, rather, what edge case is this guarding from?

canarslan12: * site removes parent of video element and saves it to anywhere (not DOM)
* therefore mutation.removedNodes not working
* when new video tag created, it uses previous one which has vscid but without vsc-controller
this code part detects this situation and it is not site specific. if there is no controller, create controller normally

igrigorik: > 2\. therefore mutation.removedNodes not working

[checkForVideo](https://github.com/igrigorik/videospeed/blob/a687bb1525e30b9fef6c2eb71f2b2d228757334d/inject.js) traverses all the children node, so it should detect wrapped video elements, no?

Also, there can be multiple vsc-controllers on the page, right? Hence my confusion for what this clause is checking: it verifies that there is at least one controller but that doesn't mean that's the "correct" one?

canarslan12: I think almost the same but it does not detect deletion for some reason.
Yes there can be multiple videos/vsc-controller. Yes it check if it's at least one on page

igrigorik: Can we try and track down why the node is not being detected as removed? The relevant logic is:

https://github.com/igrigorik/videospeed/blob/a687bb1525e30b9fef6c2eb71f2b2d228757334d/inject.js#L340-L354

vultix: @igrigorik I've looked into this.  Plex is clearing the video element's class list before removing the element.  Because of this, the above code block is being skipped.  Instead of checking if it has that class, could we instead check if it has a vscid dataset?  That isn't being removed from the node, and should continue to work.

igrigorik: @vultix yep, I think that's what we landed on here: https://github.com/igrigorik/videospeed/pull/434/files

PR #406, opened 2018-12-15
YouTube toggling Full Screen bug fix v2

canarslan12:
I disliked that idea and used this one instead. it will save previous speed and use it in new one.
it causes some flickering but it is not important


igrigorik:
Closing, resolved in #447

PR #407, opened 2018-12-15
Add speed controler for audio(basic) and Youtube fullscreen bug fix

canarslan12:
there will be a new button in settings page to enable it.
for audio element there will be one button at top of page
it is not working on spotify or soundcould because they are not using audio tags.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -25,6 +25,7 @@ chrome.runtime.sendMessage({}, function(response) {
 
       displayKeyCode: 86,   // default: V
       rememberSpeed: false, // default: false
+      audioBoolean: false, // default: false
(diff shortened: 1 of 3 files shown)


igrigorik:
@canarslan12 let's please keep the feature PR's separate from bug fixes, it makes it much harder to review and merge both. In this case, we should separate the YT fix from audio support. 

Re, audio: which sites are you testing this on?

canarslan12:
If you can merge #406, I can make this one again.
I only tested on https://radyofenomen.com which uses lazy loaded audio tags. this site also has youtube iframes which works just fine

chirag127:
is it a known issue that extension do not work on spotify.

PR #408, opened 2018-12-15
Add speed controler for audio(basic)

canarslan12:
there will be a new button in settings page to enable it.
for audio element there will be one button at top of page
it is not working on spotify or soundcould because they are not using audio tags.

I also test it in https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_audio

Note: Run button does not trigger vsc-controller for some reason but it was already bugged in previous versions.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -25,6 +25,7 @@ chrome.runtime.sendMessage({}, function(response) {
 
       displayKeyCode: 86,   // default: V
       rememberSpeed: false, // default: false
+      audioBoolean: false, // default: false
       startHidden: false,   // default: false
       keyBindings: [],
       blacklist: `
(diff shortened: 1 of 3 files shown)


jacobcolbert:
I think this should be merged. I've been using my own version (practically identical) for close to a year and can remember only a couple times that I think it behaved weirdly and put a controller somewhere it shouldn't have.

It's good to have it as a setting to turn on so it doesn't detract from the original purpose, but this is very much useful enough to merit a setting.

igrigorik:
This PR needs to be rebased to address the conflicts. Also, can we decouple `previous_speed` bits?

jacobcolbert:
For me to take care of this do I just need to make a new pull request since I can't edit this one? Would be nice to take care of this so all the recent bug fixes can be tested together and pushed out. It looks like @canarslan12 hasn't been active in a while.

igrigorik:
Closing, superseded by https://github.com/igrigorik/videospeed/pull/454.

PR #447, opened 2019-04-16
Fix issue where youtube resets speed when fullscreen

### LINKED ISSUE #370: Youtube speed resets every time while toggling Full Screen and back.

jacobcolbert:
Fixes #370

Track the speed of a video based on its source at time of last
ratechange. Newly initialize video controller with same speed as
other controllers of video with same source automatically.

Allow different speeds for different sources in same frame. Make a
controller button press only affect that controller.

Remove the deprecated settings.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -118,6 +118,14 @@
       this.initializeControls();
 
       target.addEventListener('play', function(event) {
+		if (!tc.settings.rememberSpeed) {
+	      if (!tc.settings.speeds[target.src]) {
+			tc.settings.speeds[target.src] = 1.0;
+		  }
+          setKeyBindings("reset", getKeyBindings("fast")); // resetSpeed = fastSpeed
+        } else{
+		  tc.settings.speeds[target.src]=tc.settings.speed;
+	    }	  
         target.playbackRate = tc.settings.speeds[target.src];
       });
 
--- a/inject.js
+++ b/inject.js
@@ -398,11 +398,17 @@
   function runAction(action, document, value, e) {
     var videoTags = document.getElementsByTagName('video');
     videoTags.forEach = Array.prototype.forEach;
-
+	if (e){
+		var targetController = e.target.getRootNode().host;
+	}
+	
     videoTags.forEach(function(v) {
       var id = v.dataset['vscid'];
+	  
       var controller = document.querySelector(`div[data-vscid="${id}"]`);
-
+	  if (e && !(targetController == controller)) {
+		  return;
+	  }
       showController(controller);
 
       if (!v.classList.contains('vsc-cancelled')) {
--- a/inject.js
+++ b/inject.js
@@ -1,6 +1,7 @@
   var tc = {
     settings: {
       speed: 1.0,           // default 1x
+	  speeds: {},           // empty object to hold speed for each source
 
       /**
        * these are not used and deprecated, will be removed in next update
@@ -21,7 +22,7 @@
        * these(above) are not used and deprecated, will be removed in next update
        * but should be stay there because chrome.storage.sync.get needs them.
        */
-
+ 	  
       displayKeyCode: 86,   // default: V
       rememberSpeed: false, // default: false
       startHidden: false,   // default: false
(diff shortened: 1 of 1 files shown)


alystair:
I have a feeling the original creator has stopped working on the extension, has never responded to people on GitHub - maybe worth while creating a new extension?

jacobcolbert:
I don't think that is true, he pushed out a release a month or so ago. Probably busy with some other important things though from the looks of his profile. But I def recommend downloading my version and running it while we wait because the problem drove me up the wall!

I just noticed that @canarslan12 has a pull request to fix this too using a different method. Might help if we can reach a consensus ourselves about it. I prefer my approach as it solves two birds with one stone (allows different speeds on different videos on same page). It also moves us in the direction of src based controllers that will stop double controllers on empty video containers. Do you see any problems with my solution?

inject.js
```diff
@@ -142,14 +127,14 @@
         if (event.target.readyState > 0) {
           var speed = this.getSpeed();
           this.speedIndicator.textContent = speed;
-          tc.settings.speed = speed;
```
igrigorik: Wouldn't this create an implicit history of every video you watch, and grow in an unbounded manner? The former is a non starter from a user privacy perspective.

inject.js
```diff
@@ -393,11 +378,17 @@
   function runAction(action, document, value, e) {
     var videoTags = document.getElementsByTagName('video');
     videoTags.forEach = Array.prototype.forEach;
-
+	if (e){
```
igrigorik: Can you elaborate on what this logic is doing?

inject.js
jacobcolbert: The Map containing the speeds for each video sources is never saved anywhere, and would disappear whenever the user loads a new site or closes the tab.

The variable tc.settings.speed is just the latest speed and is stored for people who have "remember speed" checked.

inject.js
jacobcolbert: When the controller buttons are pressed it passes the event e to runAction. If runAction is called by a keyBinding than e is null and it will skip the following tests and change all video speeds.

We use the event to find the videoTag owning the controller. Rather than changing the speed of all videos we check each one to see if it has the same controller and only proceed if it does.

inject.js
igrigorik: 👍

inject.js
igrigorik: Can you please add a comment to capture this in the source as well?

inject.js
```diff
@@ -127,13 +108,25 @@
...
       this.id = Math.random().toString(36).substr(2, 9);
       if (!tc.settings.rememberSpeed) {
-        tc.settings.speed = 1.0;
+		if (!tc.settings.speeds[target.src]) {
+			tc.settings.speeds[target.src] = 1.0;
+		}
         setKeyBindings("reset", getKeyBindings("fast")); // resetSpeed = fastSpeed
-      }
+      } else{
+		  tc.settings.speeds[target.src]=tc.settings.speed;
```
igrigorik: formatting: add spaces around `=`.

jacobcolbert:
Think that should all be better. Did not realize notepad++ was using tabs. Need to look into that...

garyking:
I can't wait for this update! Any ETA for the next release?

Issue #451, opened 2019-04-18
Does not work on dailymotion


igrigorik:
Example: https://www.dailymotion.com/video/x75y5s3?playlist=x63vyg

jacobcolbert:
It is somewhere cleaning out the controller div and then crashing in runAction when it calls showController on null.

igrigorik:
Closing, #452.

PR #452, opened 2019-04-18
Check that controller still exists before showing

jacobcolbert:
Partial fix for #451

Doesn't make controller stay around, but lets the hotkeys work.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -397,8 +397,11 @@
     videoTags.forEach(function(v) {
       var id = v.dataset['vscid'];
       var controller = document.querySelector(`div[data-vscid="${id}"]`);
-
-      showController(controller);
+	  
+	  if(controller){
+		showController(controller);
+	  }
+      
 
       if (!v.classList.contains('vsc-cancelled')) {
         if (action === 'rewind') {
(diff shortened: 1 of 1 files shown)


jacobcolbert:
@igrigorik Do you have some concern about this? It seems like a good change to me. The visible controller being removed should not crash out the whole extension.

inject.js
```diff
@@ -401,7 +401,9 @@
       if (e && !(targetController == controller)) {
         return;
       }
-      showController(controller);
+      if(controller){
```
igrigorik: Hmm.. if the controller is not found, shouldn't we simply return and exit out of this function? Why would we allow the rest of the code to run here? Concretely, instead of adding this check, should we add a check in the if above and exit if controller is null?

jacobcolbert: This way allows it to work on websites (like your dailymotion example, but there are many more) that clean the DOM and remove the vsc div but otherwise don't interfere.

igrigorik: Torn by this, as it's unexpected and semi broken behavior.

jacobcolbert: I understand your concerns now. Up to you. I personally think it is better than not, since it maintains the core functionality. I don't see it as a hack, but as a method to fail minimally.

If you prefer not let me know (or make the change yourself) and we can just exit. Crashing out is definitely not the best way to fail.

Issue #457, opened 2019-04-22
Speed resets when video quality changes


inian:
When the "Remember playback speed" option is unchecked, and when the quality of the video on youtube changes (due to changing network conditions, etc), the video playback speed resets to 1x each time. Does not happen when remember playback speed option is checked, but with the latest update, there is [this issue](https://github.com/igrigorik/videospeed/issues/455#issuecomment-485296516). 

Not sure, if this is the expected behaviour, but I would expect the video speed to remain the same (not reset to 1x) even when the remember playback speed option is unchecked.

jacobcolbert:
Did this happen before the latest update?

jacobcolbert:
I actually can't replicate this bug on any version. Can you say if it happened both before and after today's update?

igrigorik:
@jacobcolbert found a way to reproduce this.. Go on youtube, pick any video, start playing and adjust the speed and then hit the gear icon and select a different quality level: it resets the speed back to last remembered speed when you do this. This is a manual trigger, but I hit the same case when YT auto switches due to network quality.

PR #458, opened 2019-04-22
Update the tc.settings.speed on rateChange (Fix #455)

### LINKED ISSUE #455: Speed reverts when skipping ahead
It started today. If I load up a video it remembers my speed as it always did. If I change the speed and skip ahead in the video the speed reverts to what it was when I loaded the video.

jacobcolbert:
Fixes #455

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -1,6 +1,6 @@
   var tc = {
     settings: {
-      speed: 1.0,           // default 1x
+      lastSpeed: 1.0,           // default 1x
       speeds: {},           // empty object to hold speed for each source
 
       displayKeyCode: 86,   // default: V
@@ -76,7 +76,7 @@
         blacklist: tc.settings.blacklist.replace(/^\s+|\s+$/gm, '')
       });
     }
-    tc.settings.speed = Number(storage.speed);
+    tc.settings.lastSpeed = Number(storage.lastSpeed);
     tc.settings.displayKeyCode = Number(storage.displayKeyCode);
     tc.settings.rememberSpeed = Boolean(storage.rememberSpeed);
     tc.settings.audioBoolean = Boolean(storage.audioBoolean);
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -140,6 +140,7 @@
           var speed = this.getSpeed();
           this.speedIndicator.textContent = speed;
           tc.settings.speeds[this.video.src] = speed;
+          tc.settings.speed = speed;
```
igrigorik: What's your theory or thinking behind this one? I similarly haven't been able to reproduce this bug..

jacobcolbert: I can reproduce the bug specifically mentioned in #455, but not the potentially related #457 or #459. It is caused by line 131 in the event listener for 'play.' I added that preamble to initialize the speed in case it had not yet been initialized. But in changing to the Map of speeds for each source I removed all updating of tc.settings.speed. This updates tc.settings.speed always to the last speed.

I feel like we should refactor settings.speed to settings.lastSpeed since that is now the remaining meaning of it. I can go ahead and do that if you agree.

jacobcolbert: To confirm, did you publish the current master (through 5fec357) the other day? Are these bugs in the latest code or something else?

igrigorik: Yes, latest release includes the audio PR.

Let's refactor this, I don't think it makes sense to maintain both — only adds to confusion.

inject.js
```diff
@@ -1,6 +1,6 @@
   var tc = {
     settings: {
-      speed: 1.0,           // default 1x
+      lastSpeed: 1.0,           // default 1x
```
igrigorik: Hmm, why do we need to change the name?

jacobcolbert: With the new dict speeds, the remaining purpose of the variable previously called speed is to track the most recent speed for the purposes of people who have "remember your speed" checked. It is confusing to have a variable named speed and speeds so I thought we should rename the variable speed to better capture its current purpose.

igrigorik: Gotcha, makes sense.

PR #463, opened 2019-04-25
Let each controller track its own speed (fix #457)

### LINKED ISSUE #457: Speed resets when video quality changes

jacobcolbert:
Fixes #457 

Prevent source changes from changing speed in same video container.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -110,10 +110,11 @@
       this.parent = target.parentElement || parent;
       this.document = target.ownerDocument;
       this.id = Math.random().toString(36).substr(2, 9);
+      this.speed = 1.0;
 
       if (!tc.settings.rememberSpeed) {
         if (!tc.settings.speeds[target.src]) {
-          tc.settings.speeds[target.src] = 1.0;
+          tc.settings.speeds[target.src] = this.speed;
         }
         setKeyBindings("reset", getKeyBindings("fast")); // resetSpeed = fastSpeed
       } else {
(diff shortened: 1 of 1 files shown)


jacobcolbert:
Unfortunately after getting rid of a variable called speed I have had to reintroduce a variable called speed! But I think it is necessary. Tracking current speeds by source has solved some problems, but created #457 because the source changes despite being the same video. Each controller now can track its own speed and if the source of the video tag is changed it will pick up with the same speed.

igrigorik:
Wait, doesn't this bring us back to where we started? As in, what's the point or value in having the track by source logic if we're back to controller+video pair be the master?

jacobcolbert:
The master is still the speeds[src], but if that is not defined then it takes the speed from the controller. 

The global dict solves the problem of the same source being continuously passed between newly created video tags. (The youtube fullscreen issue #370)

Giving each controller a speed field solves the problem that this creates when a video tag has its source switched (#457)

igrigorik:
Ok, yeah, that makes sense.

PR #467, opened 2019-04-29
Hide controller for videos with no source

### LINKED ISSUE #236: Controller lifetime should be tied to src on video (no src, no controller)

jacobcolbert:
Fixes #236 

Creates a mutation observer that will detect when source changes. Hides visible controller when there is no source present. Prevents double/more controllers being visible.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -154,7 +154,7 @@
         }
       }.bind(this));
 
-      var observer=new MutationObserver((mutations)=> {
+      var observer=new MutationObserver((mutations) => {
         mutations.forEach((mutation) => {
           if (mutation.type === 'attributes' && mutation.attributeName === 'src'){
             var controller = document.querySelector(`div[data-vscid="${this.id}"]`);
@@ -163,10 +163,8 @@
             }
             if (!mutation.target.src) {
               controller.classList.add('vsc-nosource');
-              console.log('Hiding: no source');
             } else {
               controller.classList.remove('vsc-nosource');
-              console.log('showing: '+mutation.target.src);
             }
           }
         });
--- a/inject.js
+++ b/inject.js
@@ -165,7 +165,7 @@
               controller.classList.add('vsc-nosource');
               console.log('Hiding: no source');
             } else {
-              ontroller.classList.remove('vsc-nosource');
+              controller.classList.remove('vsc-nosource');
               console.log('showing: '+mutation.target.src);
             }
           }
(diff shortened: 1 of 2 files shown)


inject.js
```diff
@@ -152,7 +154,26 @@
...
+      var observer=new MutationObserver((mutations)=> {
+        mutations.forEach((mutation) => {
+          if (mutation.type === 'attributes' && mutation.attributeName === 'src'){
+            var controller = document.querySelector(`div[data-vscid="${this.id}"]`);
+            if(!controller){
+              return;
+            }
+            if (!mutation.target.src) {
+              controller.classList.add('vsc-nosource');
+              console.log('Hiding: no source');
```
igrigorik: Are the console.log's necessary? 

Sidenote: we should probably keep them around, as they can be helpful in debugging, but we should probably make a helper function (separate PR) that checks for a debug flag or some such, that can be set in development but does not otherwise need to fire for all users.

inject.js
```diff
@@ -123,6 +123,8 @@
       } else {
         tc.settings.speeds[target.src] = tc.settings.lastSpeed;
       }
+      target.playbackRate = tc.settings.speeds[target.src];
```
igrigorik: Why is this necessary in context of this PR?

jacobcolbert: This is just reorganization of line 155 moved up to here. It makes sense to have all speed initialization things together and all Listeners grouped together.

jacobcolbert: With the big new mutation block it seemed even more out of place

inject.js
```diff
@@ -152,7 +154,26 @@
         }
       }.bind(this));
 
-      target.playbackRate = tc.settings.speeds[target.src];
+      var observer=new MutationObserver((mutations)=> {
```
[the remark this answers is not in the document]
jacobcolbert: working on it! I come from eschewing all spaces, so still sometimes slip...

inject.js
jacobcolbert: Logs clearly not necessary. Forgot to take them out from  debugging and was gonna commit again to remove them but thought maybe they were worth keeping as they shouldn't fire too often. Probably not worth keeping without  a debug logger function.

inject.js
igrigorik: 👍

PR #469, opened 2019-04-29
Fully remove videoController on video removal

### LINKED ISSUE #468: videoController objects are never fully removed
Any time a video is removed VSC removes the visible div from the DOM and removes the vscid from the video tag, but never unregisters listeners. This can be seen easily in youtube by refreshing any video. Over half the time the page will "delete" an ancestor of the video and immediately recreate it. Thus multiple controllers are created for the video (generally 2). If you change the speed you can see that the logging is doubled.

jacobcolbert:
Fixes #468 

Often times video tags that are detected as "removed" are never actually removed. Even if they are, they may be placed back in the DOM later. In master either of these will result in multiple eventListeners and videoController js objects for that video. This PR prevents that by removing the eventListeners and giving videos a reference to their videoController object.

It also increases efficiency by checking that removed objects are not present in the DOM before recursively checking for removed videos. According to devtools this accounts for ~half of time VSC currently uses on youtube loads.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -346,7 +346,8 @@
 
 
       function checkForVideo(node, parent, added) {
-        if (document.body.contains(node)) {
+        // Only proceed with supposed removal if node is missing from DOM
+        if (!added && document.body.contains(node)) {
           return;
         }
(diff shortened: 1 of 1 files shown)


igrigorik:
@jacobcolbert not directly related to this PR, but could you please ping me at ilya-at-grigorik.com over email?

PR #471, opened 2019-05-05
Add feature to jump to marker

### LINKED ISSUE #470: Feature request: marker for specific start time
I don't really program or anything, and I made a github account just for this feature request, so my apologies if I'm doing anything against etiquette here. 

So I usually use this extension to transcribe music off of YouTube, and adding a feature that would allow one to add a specific marker of time to jump back to a specific point in the video would be really useful.

For example if there were a section in a video at 0:36 seconds that I wanted to transcribe, I could set a marker with a hotkey and then press a different hotkey to automatically jump back to 0:36 in the video at the same speed that was already set. Hope this makes sense, and thanks for the extension!

jacobcolbert:
Closes #470 

Adds two options for new custom hotkeys. One to set a marker and another to jump to a marker.

Also refactors logic so that list of actions with no value only needs to be specified once.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -484,6 +484,10 @@
           pauseSpeed(v, value);
         } else if (action === 'muted') {
           muted(v, value);
+        } else if (action === 'mark') {
+          setMark(v);
+        } else if (action === 'jump') {
+          jumpToMark(v);
         }
       }
     });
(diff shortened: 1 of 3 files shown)


Ayoitswiz:
@igrigorik 
The [readme.md](https://github.com/igrigorik/videospeed/blob/master/README.md) file has not been updated for this key-binding. I'd add it myself but I'm unaware of it's function, which is why I'm leaving this comment.

igrigorik:
Great catch, yes, we should document it..

PR #472, opened 2019-05-05
Fix whitespace issue in blacklist

### LINKED ISSUE #436: Adding empty line in blacklisted sites merges two lines together
If my blacklisted sites is like this, with an empty line for organizational purposes:

```
reddit.com
twitter.com

twitch.tv
```

Then when I save it and re-open the settings page, the value is now:

```
reddit.com
twitter.comtwitch.tv
```

When they should either stay as I entered them, or at least don't merge the two lines together.

jacobcolbert:
Fixes #436 

Change regex to not capture newlines

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -1,5 +1,5 @@
   var regStrip=/^[\r\t\f\v ]+|[\r\t\f\v ]+$/gm;
-  
+
   var tc = {
     settings: {
       lastSpeed: 1.0,       // default 1x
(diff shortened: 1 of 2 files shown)


garyking:
It seems like the bug still exists. Has this been pushed to the Chrome extension yet?

igrigorik:
Hmm, nope. Thanks for the nudge, on its way now.

garyking:
After this update, some of my blacklisted websites disappeared.

Issue #428, opened 2019-02-01
Preferred speed changes to 0 on pause, requiring reset speed key press


eggbrow:
Have been using this now for 2 years. It's so useful that i could not cope without it. As of mid december i've been having an issue where (i'm fairly sure) it changes preferred speed to 0 after every time i pause. I then have to press r to reset that to 1. Anybody else getting this issue?
![image](https://user-images.githubusercontent.com/25536683/52122267-4dee9f80-261a-11e9-95b1-aed2291cf76c.png)

eggbrow:
i am on
![image](https://user-images.githubusercontent.com/25536683/52122292-6d85c800-261a-11e9-97a1-429a23a1f35d.png)
windows 8.1 64bit

and extension settings are set to
![image](https://user-images.githubusercontent.com/25536683/52122312-7f676b00-261a-11e9-81c6-df9d4181d7a0.png)

eggbrow:
I removed the additional pause shortcut (was for sites where the spacebar didnt pause by default or with player focused) and everything working properly. I'll leave this issue here in case anybody else has same problem.

igrigorik:
Thanks for flagging this. This is a bug, we shouldn't reset the speed on pause; unpausing should reset back to original speed.

/cc @canarslan12

igrigorik:
Should be resolved via #475.

PR #475, opened 2019-05-18
Use built in pause() and play()

### LINKED ISSUE #474: Youtube bug (not tested other platforms)
I have the "remember playback speed checked" but this bug happens with that checked off as well. I usually watch everything at ~2x speed
Google Chrome is up to date
Version 74.0.3729.157 (Official Build) (64-bit)

When on YT and I hit K (youtube native for pause), and then K again - everything works fine.

However, when I hit 'space' it pauses, but reverts the speed to 0.00 and the video shows it's playing, but it's stopped. So I either have to hit R or G to have it play again.

PS.  In my settings my preferred speed is set to 2.0 but G always places it to 1.8 :((

jacobcolbert:
Fixes #474 and #428

I feel like I must be missing the history of why this was implemented the previous way. But this has been working for me so far fine.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -483,7 +483,7 @@
         } else if (action === 'fast') {
           resetSpeed(v, value);
         } else if (action === 'pause') {
-          pauseSpeed(v, value);
+          pause(v);
         } else if (action === 'muted') {
           muted(v, value);
         } else if (action === 'mark') {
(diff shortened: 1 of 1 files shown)


jacobcolbert:
Ok, I guess I found out why it was implemented that way maybe? The behavior if you set a VSC pause hotkey to the same thing as the native pause hotkey depends on the player. It works correctly on youtube now, but on amazon player it will end up toggling twice and end back in the state it started. Not really sure what to do about that because it seems it will depend on whose event listener executes first.

igrigorik:
Unfortunately this is not a game we can ever win. There are a few more tricks we could play to take precedence (i.e. register our listener earlier than others) but even there, stopping the event from propagating might break some functionality that you need or want on the site. 

I think the best option and guidance we have here is to instruct the user to try and pick a key that won't conflict.. Of course, depending on what site you use, that may differ.

Issue #362, opened 2018-07-18
Suggestion: Option to select the opacity of the controller


shmuel1234:
It would be really nice to have the option to select the opacity of the controller.

igrigorik:
Thanks for the suggestion. No immediate plans to tackle this, but added it to the wishlist queue.

PR #478, opened 2019-05-21
Add feature to change controller opacity

GabrielM98:
I have found that sometimes the controller is hard to see against certain backgrounds. Hence I have implemented a feature to allow the user to change the opacity of the controller from the options page (as shown in the image below).

Screenshot from 2019-05-21 23-46-28

This feature had been requested in #362.


igrigorik:
Thanks for the contribution!

PR #495, opened 2019-07-17
Add teams.microsoft.com to default blacklist

lencioni:
Microsoft Teams is used for video conferencing, and this extension
causes all of the videos to be half cut off. I think it would be a good
idea to add it to the default blacklist.


igrigorik:
thank you!

PR #527, opened 2019-10-03
Update manifest.json

kokikohara:


jacobcolbert:
Pretty sure this is out of date solution. Do you have any reason to believe it is not?

https://stackoverflow.com/questions/47757849/chrome-extension-disabled-this-extension-may-have-been-corrupted-chrome-63-0-3#comment100832620_47763054

PR #528, opened 2019-10-05
Add ability to enable/disable from toolbar

thewheat:
Addresses https://github.com/igrigorik/videospeed/issues/464 by adding the ability to enable/disable VSC from the toolbar popup

Hope this works but happy to take on any feedback 👍

**Default view or when VSC is enabled**
image

**On enabling/disabling**
Will show a messaging asking to reload 
image

**If disabled it will show a "Enabled" button instead**
image

**Also added to options/settings page**
image

PR #529, opened 2019-10-05
Use .currentSrc instead of .src to correctly identify video (Fix #506)

### LINKED ISSUE #506: No controls when viewing video directly
Hi, I noticed recently that when visiting videos with a direct URL, i.e., the URL shown in the address bar is some https://.../video.mp4, the controls no longer show up even when triggered.
However, the shortcuts to increase/decrease the speed are still working.

When debugging, I noticed that this condition is true:https://github.com/igrigorik/videospeed/blob/5641369385cf4175e308c8e188c9bb70226a12da/inject.js#L206
which inserts the `vsc-nosource` class and effectively hides the controls.
I'm not sure why this.video.src is == "" while the video is effectively playing, but so that breaks the controls.

Despite this code being introduced 4 months ago, I have seen the controls on videos watched directly in the last few months, so it's possibly a browser change.

thewheat:
Fixes https://github.com/igrigorik/videospeed/issues/506

- When loading a local file `video.src` is not defined and that causes the controller to be hidden due to [this PR](https://github.com/igrigorik/videospeed/pull/467)
- `video.currentSrc` is set and should be the attribute to use to refer to the existing src as there could be multiple sources (see below for details) 
- so changing references of `video.src` to `video.currentSrc`

image

Videos can be specified with no direct `src` attribute but multiple `source`
- https://www.w3schools.com/html/html5_video.asp
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video
image

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -122,28 +122,28 @@
       this.speed = 1.0;
 
       if (!tc.settings.rememberSpeed) {
-        if (!tc.settings.speeds[target.src]) {
-          tc.settings.speeds[target.src] = this.speed;
+        if (!tc.settings.speeds[target.currentSrc]) {
+          tc.settings.speeds[target.currentSrc] = this.speed;
         }
         setKeyBindings("reset", getKeyBindings("fast")); // resetSpeed = fastSpeed
       } else {
-        tc.settings.speeds[target.src] = tc.settings.lastSpeed;
+        tc.settings.speeds[target.currentSrc] = tc.settings.lastSpeed;
       }
 
-      target.playbackRate = tc.settings.speeds[target.src];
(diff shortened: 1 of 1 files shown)


jondcoleman:
When will this release be available?

igrigorik:
@jondcoleman thanks for the nudge :-)

@jacobcolbert any outstanding issues or PRs on your side that you'd want to see landed before we ship new release?

jacobcolbert:
#528 has been working perfectly for me. The code looks fine. But not sure if it follows all chrome extension best practices though if you wanna take a look.

garyking:
@igrigorik Please release a new version of this extension, that includes this PR. Thanks!

Issue #526, opened 2019-10-02
Disable website keybinds for viewing controller


Skarrum:
So I found a keybind layout I enjoy with num0 showing and hiding the controller, num1-9 setting the speed from 1 to 9 percent sequentially, and num+ and num- changing it by .10 speed in either direction. On all the other options you can disable website keybinds for each individual action, but the same doesn't apply to the show/hide controller option. For example, since I can't disable that if I want to view my controller while watching Youtube, or another site with similar keybinds it will send me to the beginning of the video.

jacobcolbert:
This is an oversight. Not sure why this was implemented seperately. Will take a look soon hopefully.

thewheat:
Hope it was ok that I created a PR for this 😄  https://github.com/igrigorik/videospeed/pull/533

PR #533, opened 2019-10-11
Add ability to disable website keybindings for show/hide controller (Fixes: #526)

thewheat:
- Adding the select dropdown to allow disabling website keybinds for the "Show/hide controller" option similar to the other keybindings
- it only shows when clicking "show experimental features"
image

**Demo**
Demo of it showing only after clicking "show experimental features" and of it being able to persist on saving
show-hide

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -6,7 +6,6 @@
       speeds: {},           // empty object to hold speed for each source
 
       displayKeyCode: 86,   // default: V
-      displayForce: false,
       rememberSpeed: false, // default: false
       audioBoolean: false,  // default: false
       startHidden: false,   // default: false
@@ -75,7 +74,6 @@
         keyBindings: tc.settings.keyBindings,
         version: tc.settings.version,
         displayKeyCode: tc.settings.displayKeyCode,
-        displayForce: tc.settings.displayForce,
         rememberSpeed: tc.settings.rememberSpeed,
         audioBoolean: tc.settings.audioBoolean,
         startHidden: tc.settings.startHidden,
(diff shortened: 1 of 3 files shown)


jacobcolbert:
This looks nice, thanks for the work. But I am left wondering why show/hide is treated differently from every other feature. I would prefer a small refactor from the original to just treat show/hide like everything else. It seems like this was explicitly not done in #350. Do you have any ideas why? @igrigorik @canarslan12

igrigorik:
@jacobcolbert scratching my head as well, I think this was an oversight as I don't recall any specific reason why we decided to specifically omit this feature. Also, +1 to your recommendation on refactor.

thewheat:
Cool. By refactor just meaning to make it styled like the rest with  
- `Show/hide controller` as a form select input
- And for the non customizable input to be disabled?

image

jacobcolbert:
In terms of external behavior, yes that is exactly what I mean.

In the code I mean that show/hide is treated by itself as an exceptional case while every other feature is treated uniformly. Specifically, in options.html it is implemented as <div class="row"> rather than <div class="row customs"> like all of the following features. And in the options.js it is not included as an option in the dropdown. 

The code should treat all of these features the same, so that if we change things in the future it will automatically affect all of them equivalently.

igrigorik:
Nice! I think this looks good now, nice touch on custom bindings carryover.. 

@jacobcolbert any feedback before we merge?

jacobcolbert:
Looks great to me. Gonna fix some indentation errors that were already there. 

Very nice touch on carrying over the custom bindings. Would prolly have missed that myself. Can that code be removed in the next version after this is pushed live?

thewheat:
Hmmm that's a good question. 

I think the part in `inject.js` should be kept

https://github.com/igrigorik/videospeed/blob/29788ba82ad93c2798f26676acedf23ca662105b/inject.js#L93-L101

It is similar to the current code to initialize the other shortcuts
https://github.com/igrigorik/videospeed/blob/29788ba82ad93c2798f26676acedf23ca662105b/inject.js#L26-L71

* * * 
The `options.js` change is dependent if a `storage.keyBindings` has been saved after the update as that is when the new `display` action is set via `save_options`. 

Then we can remove this line

https://github.com/igrigorik/videospeed/blob/29788ba82ad93c2798f26676acedf23ca662105b/options.js#L187-L190

And also this following one as `item["key"]` would be defined

https://github.com/igrigorik/videospeed/blob/29788ba82ad93c2798f26676acedf23ca662105b/options.js#L197-L199

Issue #532, opened 2019-10-10
Enable use of "arrow keys" for shortcuts


habassa5:
Seems the most obvious keystroke shortcuts would use arrows keys instead of what appear to be random letters... I've tried to change the default shortcuts to utilize the arrow keys but they don't take. Please fix.

thewheat:
I do see that the code filters out only certain keypresses. I've added support for arrow keys and function keys in this PR https://github.com/igrigorik/videospeed/pull/534 if the authors are happy to support them 👍

PR #534, opened 2019-10-13
Add ability to use arrows and Function keys for shortcuts (Fixes: #532)

thewheat:
Addresses https://github.com/igrigorik/videospeed/issues/532 if this is something we'd want to support.

Also added ability to recognize Function keys (e.g. F1...F12)


igrigorik:
FWIW, we intentionally omitted arrow keys because they tend to conflict with many players, which already assign them to ff / rw and volume controls. That said, I don't feel strongly about this.. I suspect that folks just might find it irritating because that's not what they expect.

We can land this and see if and what the feedback is. :)

PR #538, opened 2019-10-30
Adding feature - Show the controller briefly when adjusting speed and the controller is set to hidden.

EonYang:
Hey, community,

As a loyal user of this extension, I love it so much and use it thousands of times every day.

I like to hide the controller to have a clear view of the screen. However, when I adjust the video speed, I feel anxious because I can't see the current speed since the controller is hidden. 

I don't like this uncertainty, so I added a small feature to this extension, and hope you guys like it.

When the rate changes and the controller is set to hidden, the controller will show up briefly to give the user clear feedback, and then disappear again. 

To do this, I added a new action called `blink` and call `runAction('blink')` in the event `'ratechange'`.

vs

In this demo, the controller is hidden by default and I only pressed `d` and `s` several times.

The default duration is set to 1000 ms, but you can adjust it by passing `value` into `runAction()`, in case we want to make it customizable in the future.

Please let me know your thoughts, comments or feedback. 

That's it, thank you guys!

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -172,6 +172,7 @@
           chrome.storage.sync.set({'lastSpeed': speed}, function() {
             console.log('Speed setting saved: ' + speed);
           });
+          // show the controller for 1000ms if it's hidden.
           runAction('blink', document, 1000, null);
         }
       }.bind(this));
@@ -496,6 +497,7 @@
           controller.classList.add('vsc-manual');
           controller.classList.toggle('vsc-hidden');
         } else if (action === 'blink') {
+            // if vsc is hidden, show it briefly to give the use visual feedback that the action is excuted.
             if(controller.classList.contains('vsc-hidden') || controller.blinkTimeOut !== undefined){
               clearTimeout(controller.blinkTimeOut);
               controller.classList.remove('vsc-hidden');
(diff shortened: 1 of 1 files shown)


EonYang:
> Very nice! I think this makes sense and is a good improvement.

Thank you! I'm glad that you like it. 

I did more tests and fixed a bug in the new commit.

EonYang:
Occasionally, I encounter an issue that the VSC has a class `vsc-nosource` on youtube.

I tested and can confirm that this issue exists in the current release.

Though I made my code work by removing `vsc-nosource` manually, we might wanna open an issue and discuss it thoroughly.

ChadBailey:
I've been testing this (albeit, not particularly thoroughly) on my machine. It has been behaving as expected for me. I really like the enhancement, thanks so much for doing this!

The only suggestion I could think of is some users may want it to blink on load or hover, in order to know which elements VSC is able to control. That isn't to say I'm suggesting you make this change though

jacobcolbert:
Also think it looks really nice. 

I'm not happy about the removing of the vsc-nosource tags, they are there for a reason. I'd prefer a merge with before commit b7c4f2fe94808ac2552e5a86034b78520060d31f.

If there is a nosource issue which i think has cropped up in some places I agree we should discuss it and fix it separately from this bugfix.

EonYang:
Yeah, I agree. It's not a good idea to tamper the noreseorce tag in this PR.

Issue #516, opened 2019-09-14
Video Speed Controller UI overtakes fixed navbars on some sites


gruiz17:
Probably not a world-ending issue, but still an interesting visual quirk -- when some video streaming sites have a fixed navbar at the top, and you scroll down, the UI of VSC scrolls such that it appears on _top_ of the navbar.

Here's an example of what I mean by title, on Rakuten Viki:

![uiBug](https://user-images.githubusercontent.com/2757082/64905274-90855980-d68a-11e9-8256-9c2bfde760e0.gif)

I'm guessing this is borne of the fact that `z-index` on `#controller` in `shadow.css` [is set to 9999999](https://github.com/igrigorik/videospeed/blob/master/shadow.css#L24).

Interestingly, this doesn't happen on YouTube, even though YouTube's navbar `z-index` is set to 2020.

If possible, what ways are there to make sure there's some consistency to the scrolling behavior with VSC wrt sites with fixed navbars?

---

Steps for reproduction:

1. With VSC, Go to a video on Viki
2. Scroll down

ChadBailey:
Great bug report, thanks for the detail! I went ahead and took a stab at fixing this one since it seems simple enough. I've tested my fix on as many sites as I could think of, but if you know of any other examples please do share.

ChadBailey:
I was able to fix this bug in https://github.com/igrigorik/videospeed/pull/539 but I realized this will probably be the lesser evil than having some situations where the controller is unreachable. For this reason, I closed my request and it was not merged in.

ChadBailey:
I'm labeling this as won't fix and closing because this behavior is preferred to not being able to access the controller at all. Still hoping one day we can come up with a clever way to do both

Issue #503, opened 2019-08-08
Stack to the back layer on Youtube Miniplayer


franciscocorrales:
Buttons are stack on the back, being unreachable.
On the Youtube Miniplayer visualization. 

While on the miniplayer, I can't change the speed. I have to set the default size, change the speed, and then change again to miniplayer.

<img width="379" alt="Screen Shot 2019-08-08 at 10 49 07 AM" src="https://user-images.githubusercontent.com/4010693/62721854-5e694380-b9ca-11e9-8d9e-efdf9170d3b8.png">

![evidence](https://user-images.githubusercontent.com/4010693/62721925-8bb5f180-b9ca-11e9-979a-9b6280e8fc6f.gif)

Google Chrome: Version 76.0.3809.100 (Official Build) (64-bit)

jvschiavo:
I recommend using the hotkeys, I just use S and D to change speed, it is much faster! :)

ChadBailey:
This bug appears to be related to #516 that I'm working on. Thanks for the report, hoping my current logic works with it.

PR #539, opened 2019-11-01
bugfix for #516, sets zindex to 1 higher than video element

ChadBailey:
Please do scrutinize this, JavaScript isn't my strongest suit if I'm honest.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -124,12 +124,12 @@
     // I don't _expect_ any errors here, this is a defensive coding practice.
     // This try/catch block should be removed if debugging this function
     try {
-      i = target.style.zIndex;
       if (target === null) {
         // In the event a recursive call reaches the top of the html tree, 
         // return default + 1, which is 1.
         return 1;
       }
+      i = target.style.zIndex;
       if (isNaN(i) === false) {
         i++;
       } else {
--- a/inject.js
+++ b/inject.js
@@ -125,7 +125,12 @@
       if (target.dataset['vscid']) {
         return target.vsc;
       }
-
+      this.zindex = target.style.zIndex;
+      if (this.zindex === "") {
+        this.zindex = 1;
+      } else {
+        this.zindex++;
+      }
       this.video = target;
       this.parent = target.parentElement || parent;
       this.document = target.ownerDocument;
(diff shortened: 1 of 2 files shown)


inject.js
```diff
@@ -125,7 +125,12 @@
       if (target.dataset['vscid']) {
         return target.vsc;
       }
-
+      this.zindex = target.style.zIndex;
+      if (this.zindex === "") {
```
igrigorik: zindex can be a string other than empty string: https://www.w3schools.com/jsref/prop_style_zindex.asp

- Would setting to 1 potentially cause issues when initial value is not empty string?
- If we set the value to "inherit" which should be the z-index of the video element, would that work?

ChadBailey: Thanks for the thoughts... I'm going to try using inherit and see how that performs. I didn't realize that was an option, thanks for linking the docs!

ChadBailey: I really like how much cleaner it is, but it seems using "inherit" is only working for Vimeo for some reason. It is not working for YouTube, LinkedIn Learning, or Viki in my quick tests.

I haven't put a lot of effort into investigating why at this point. I need to get going for now but I will revisit and reply once I've had the chance to look at it more closely.

ChadBailey: As best as I can tell, the issue with inherit not working as intended seems to stem from it using the same z-index as the element.

In turn, the final choice as to whether it's shown or not depends on what I infer is [the rules governing the way "auto" works](https://www.w3schools.com/cssref/pr_pos_z-index.asp) - "Sets the stack order equal to its parents".

I've come up with a [pretty comprehensive solution](https://github.com/igrigorik/videospeed/pull/539/commits/4ca7611529def6ebacf094c33d96892a783a55c1) that seems to be accounting for every scenario and ultimately making the z-index 1-level higher than the video. This does produce some interesting results in edge cases. For example, here is a facebook story video screenshot:

image

This twitch screenshot shows the same scenario, but with expected(?) results:
image


I don't think this is avoidable as this turns in to a matter of opinion. That is to say, how would one know when to render over or under the blocking element? In my opinion, making it 1 layer higher than the video element is an acceptable outcome, but I certainly can only speak for myself.

ChadBailey:
Based on my comments above and the fact that there are other bugs possible from choosing a lower z-index such as #503 , I think this may require a design decision before an implementation can be settled on. I'll leave this branch with the solution I made alone if you decide to use this implementation, but for now I'm closing this PR.

In spite of choosing to close the PR, I think this has been a worthwhile journey. I've certainly expanded my understanding of z-index substantially.

PR #544, opened 2019-11-19
Prevent disabling default keybindings from disabling text input.

### LINKED ISSUE #543: BUG: Youtube homepage search query acting weird when website keybindings are disabled in favor of videospeedcontroller key bindings
Lets say i am on youtube.com (the homepage) and I type in "asdfg" in the search bar.........lets also say I have keybindings set up for the extension namely "d" and "s".......upon typing this word in the search bar only "afg" shows up while s and d are cut out.......What do I do about this problem aside from blacklisting youtube or choosing not to disable key bindings for the extension....if it helps i have ublock, im using chrome and I have a mac. Please help me this bug is super annoying...I feel like the key bindings for this extension shouldn't always be enabled in certain areas especially in cases where you are typing things in a search query......please help..

jacobcolbert:
Fixes #543 

The check for text input was previously done on document.activeElement. However because VSC is running in multiple contexts (and events are passed to parent contexts since #334) the activeElement in some contexts is the body. 

This changes it to look at the event.target for text editability instead.

Test driving now for unexpected issues. Let me know what y'all think.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -364,9 +364,9 @@
           }
 
           // Ignore keydown event if typing in an input box
-          if (document.activeElement.nodeName === 'INPUT'
-              || document.activeElement.nodeName === 'TEXTAREA'
-              || document.activeElement.isContentEditable) {
+          if (event.target.nodeName === 'INPUT'
+              || event.target.nodeName === 'TEXTAREA'
+              || event.target.isContentEditable) {
             return false;
           }
 


throwaway2342:
So what happens now? Is the bug fixed?

jacobcolbert:
The bug is fixed in the master version here. If you download the folder and
load the unpacked extension you can use it now. I'd only recommend this if
you feel you urgently need this fix as you won't get updates unless you
remember to return to the Chrome store version.

Otherwise the fix will be pushed out live soon enough.

Issue #542, opened 2019-11-13
How do i blacklist the homepage of a website but not the rest of the website? 


throwaway2342:

thewheat:
Hi @throwaway2342 👋Do you have an example of the website in question?

**Scenario 1**
If the homepage is `domain.com/index.html` you could add that to the blacklist and it wouldn't match on `domain.com/page1`. 

**Scenario 2**
If this is something like `domain.com` is the homepage and the rest of the site is `domain.com/page1`, with the current implementation I don't think it is possible.

Currently the blacklist works on
- getting the current URL of the page you're on (e.g. domain.com/page1)
- and if it matches any line specified in your blacklist (e.g. domain.com)
Then it will not load
https://github.com/igrigorik/videospeed/blob/ac53d3ea2ba5ca508cc46e1295e81be36416e5af/inject.js#L288-L303

If https://github.com/igrigorik/videospeed/issues/167 is implement to allow regex in the blacklist, there could be a way to achieve Scenario 2 😃 Then using something like `domain\.com$` (depending on the implementation of the regex) would work for that scenario

throwaway2342:
Hello, 

What is a regex?  
The homepage in question is "youtube.com" 

I tried putting domain\.com$ ........in the blacklist and it still doesn't block video speed controller on the homepage in question

thewheat:
> What is a regex?
A regex is a regular expression and allows for more complex search and matching of text. Thus it would allow a more robust blacklisting solution. It is a bit technical to use so it can be a bit complicated as well 😅

> The homepage in question is "youtube.com"

Ah ok, if it is youtube I don't think this would be possible until we have an added feature that is discussed in https://github.com/igrigorik/videospeed/issues/167 to allow regex in the blacklist

> I tried putting domain.com$ ........in the blacklist and it still doesn't block video speed controller on the homepage in question

I'm afraid this won't work in the current version. We would need to add regex support to the extension first. Once that is implemented we can revisit this, unless anybody else has another way that this could be done

ChadBailey:
Fixed in https://github.com/igrigorik/videospeed/pull/546

If the changes are merged in, the way you will do this is just add `/(.+)youtube\.com(\/*)$/gi` to your blacklist.

PR #546, opened 2019-11-24
Allow regex in blacklist fixes #167 #542

### LINKED ISSUE #167: Allow regexes in blacklist
Twitch live streams are usually found at `https://www.twitch.tv/foo`, while VODs (replays of previously broadcast streams) are usually found at `https://www.twitch.tv/foo/v/id`. While it is common for me to want to speed up VODs, I never intentionally speed up or speed down live streams, yet it happens frequently when I accidentally hit a shortcut key.

If there is no way to detect whether a video is a live stream (#152) and create an option to disable speed controls if so, then I’d like to have the ability to blacklist the first URL type and whitelist the second one. Would it be possible for the blacklist to allow either plain strings or regexes (delimited by `/…/` I suppose)?

ChadBailey:
Pretty straight forward, just checks if string begins with "/" then just does a raw regex comparison if so. This does not do any validation if the regex is valid. I didn't believe that was necessary but perhaps I was mistaken

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -275,7 +275,7 @@
           // the first element of the target, which may not be the parent.
           this.parent.insertBefore(fragment, this.parent.firstChild);
       }
-	  return wrapper;
+      return wrapper;
     }
   }
 
--- a/inject.js
+++ b/inject.js
@@ -293,7 +293,11 @@
       }
 
       if (match.startsWith('/')) {
-        var regexp = new RegExp(match);
+        try {
+          var regexp = new RegExp(match);
+        } catch(err) {
+          return;
+        }
       } else {
         var regexp = new RegExp(escapeStringRegExp(match));
       }
--- a/inject.js
+++ b/inject.js
@@ -292,7 +292,12 @@
         return;
       }
 
-      var regexp = new RegExp(escapeStringRegExp(match));
+      if (match.startsWith('/')) {
+        var regexp = new RegExp(match);
+      } else {
+        var regexp = new RegExp(escapeStringRegExp(match));
+      }
+
       if (regexp.test(location.href)) {
         blacklisted = true;
         return;
--- a/options.html
+++ b/options.html
@@ -112,7 +112,7 @@
         <label for="blacklist">Blacklisted sites on which extension is disabled<br/>
             (one per line)<br/>
             <br/>
-            <em>Regex is supported. Be sure it is in "//g" format.<br/>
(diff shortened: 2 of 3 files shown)


inject.js
```diff
@@ -292,7 +292,12 @@
         return;
       }
 
-      var regexp = new RegExp(escapeStringRegExp(match));
+      if (match.startsWith('/')) {
+        var regexp = new RegExp(match);
```
igrigorik: If we don't validate and this throws an exception.. the whole init function will fail, right? I think we need to guard for this here, or validate at input time (harder, but better). 

Example here: https://stackoverflow.com/questions/17250815/how-to-check-if-the-input-string-is-a-valid-regular-expression

options.html
```diff
@@ -109,7 +109,12 @@ <h3>Other</h3>
         <input id="controllerOpacity" type="text" value="">
       </div>
       <div class="row">
-        <label for="blacklist">Blacklisted sites on which extension is disabled<br/>(one per line)</label>
+        <label for="blacklist">Blacklisted sites on which extension is disabled<br/>
+            (one per line)<br/>
+            <br/>
+            <em>Regex is supported. Be sure it is in "//g" format.<br/>
+            ie: /(.+)youtube\.com(\/*)$/gi</em>
```
igrigorik: Let's add a simple validator on the input here. If we allow bad expressions, the user will have a really hard time catching this error downstream.

ChadBailey: Makes sense, thanks for the feedback. I'll see what I can do.

inject.js
ChadBailey: You're probably right, I don't really know... I'll see if I can capture the error in the event the regex is invalid and direct it to pass to the next match.

ChadBailey: Should be fixed in commit https://github.com/igrigorik/videospeed/pull/546/commits/570c971f9a3a1d843c18b1e71b9cb410bc590f1b

options.html
ChadBailey: This is going to take a little more effort than I have time for tonight as I need to create a way to visually represent a failing regex and prevent submitting with a failure. I'll try and get something here within the next couple of days.

ChadBailey: Nevermind! I didn't notice before that there was already some signaling back to the user when they click "Save", so I just hijacked that. Seems like it's working https://github.com/ChadBailey/videospeed/commit/ac8b175a9870e56c7103df6f7637aa7b043baf66

Issue #521, opened 2019-09-24
not working on pluralsight.com


salsamaani:
thanks alot, it is a usefull great extention
but since a week i cant apply it on pluralsight.com (i am using chrome)

jacobcolbert:
Can you find a video that doesn't require creating an account i could test on?

thewheat:
This video can be accessed without a login and shows the problem
https://app.pluralsight.com/course-player?clipId=1e8b695e-a148-4b6b-83d0-07c9ddbb9b73

![image](https://user-images.githubusercontent.com/892961/66249565-e28c3d00-e767-11e9-9bac-9bfb721e1246.png)

Could probably be something to do with their own speed controls
![image](https://user-images.githubusercontent.com/892961/66249613-c9d05700-e768-11e9-898c-e985bd985975.png)

Strange enough this link allows the video speed to be maintained https://app.pluralsight.com/player?name=88240101-1955-4c4c-8404-653a2fc3f135&mode=live&clip=0&course=aws-security-introduction&author=jordan-yankovich 
- from the URL path it uses `player` instead of `course-player`
- logged in users always get redirected to `course-player` so they can't manually go to `player`

**Demo of the problem**
![pluralsight](https://user-images.githubusercontent.com/892961/66249545-ace75400-e767-11e9-809c-238cbc299bda.gif)

thewheat:
Dug in a bit more and there is a `ratechange` event listener

![image](https://user-images.githubusercontent.com/892961/66696850-264feb00-ed02-11e9-9494-8ee8dd2234e6.png)

If you remove it, it will work (hover over it in dev tools and a "Remove" button will show to remove it)
![image](https://user-images.githubusercontent.com/892961/66696934-2997a680-ed03-11e9-8fd5-075e007ce72f.png)


Looking at some of the code it looks like there is some stored state which VSC doesn't modify and thus it resets back to the stored playbackRate value 
```
  onRateChange = rate => {
    const { current: player } = this.element

    player && this.setState({ reportedPlaybackRate: player.playbackRate })
  }
```
...
```
    if (nextProps.playbackRate !== reportedPlaybackRate) {
      this.setPlaybackRate(nextProps.playbackRate)
    }
```

ChadBailey:
Currently working on this one... I'll use this thread since it's the original and has more details. Here's my post from #531 

[...]

PR #547, opened 2019-11-25
Bugfix/521 plural sight

ChadBailey:
This adds a second "blacklist", but this list is to block video rate change event listeners on these sites. I have tried it on pluralsight and it works as expected. ~~There is one caveat, this prevents the rate from being displayed properly in the controller. Fortunately, all other functionality is maintained. Perhaps that could also be fixed by finding another more direct way to update the controller when adjusting the rate.~~

Update: Fixed!

Ignore the extraneous "a" in "Plural" in the branch name lol


ChadBailey:
Fixes https://github.com/igrigorik/videospeed/issues/521

ChadBailey:
working on resolving merge conflicts and other niceties... more soon to come :)

inject.js
```diff
@@ -19,7 +19,7 @@
         imgur.com
         teams.microsoft.com
       `.replace(regStrip,''),
-      ratechange: `
```
ChadBailey: I discovered `ratechange` to be ambiguous so I changed it to `blacklistrc` which means `blacklist rate change`

inject.js
```diff
@@ -154,22 +154,25 @@
 
       this.div=this.initializeControls();
 
-      target.addEventListener('play', this.handlePlay = function(event) {
-        if (!tc.settings.rememberSpeed) {
-          if (!tc.settings.speeds[target.currentSrc]) {
-            tc.settings.speeds[target.currentSrc] = this.speed;
+      if (isRateChangeBlocked() === false) {
```
ChadBailey: This original listener was added to fix YouTube fullscreen from resetting the playback rate. Not sure why a YT specific bug has been applied globally, but in the case of pluralsight it has the opposite effect. For now, bypassing this fix seems to work fine.

inject.js
```diff
@@ -211,10 +221,11 @@
 
     tc.videoController.prototype.initializeControls = function() {
       var document = this.document;
-      var speed = parseFloat(tc.settings.speeds[this.video.currentSrc]).toFixed(2),
```
ChadBailey: The intention of the original author was confusing to me, so I broke this out into their own declarations

inject.js
```diff
@@ -275,33 +286,75 @@
           // the first element of the target, which may not be the parent.
           this.parent.insertBefore(fragment, this.parent.firstChild);
       }
-	  return wrapper;
```
ChadBailey: Erroneous tab replaced with spaces

inject.js
```diff
@@ -275,33 +286,75 @@
...
+      return wrapper;
     }
   }
 
-  function initializeWhenReady(document) {
```
ChadBailey: This looks worse than it really is, just pushed out blacklist into its own function and likewise with blacklistrc. initializeWhenReady looks much cleaner now, and it's easy to identify if blacklisted from anywhere now.

inject.js
```diff
@@ -275,33 +286,75 @@
...
+  return blockRateChange;
+}
 
+  function initializeWhenReady(document) {
+    if (isBlacklisted()) {
+      return;
+    }
+    if (isRateChangeBlocked()) {
+      document.body.addEventListener("ratechange", function (event) {
+        event.stopImmediatePropagation();
```
ChadBailey: This is the secret sauce that makes the ratechange event listener blocker work

inject.js
```diff
@@ -346,6 +399,7 @@
         if (inIframe())
           docs.push(window.top.document);
       } catch (e) {
+        // no operation
```
ChadBailey: I didn't like an open and closed parenthesis as it made me believe something was cut from this section erroneously.

inject.js
```diff
@@ -448,6 +501,13 @@
       });
   }
 
+  function setSpeed(controller,video,speed) {
```
ChadBailey: Moved setSpeed into its own function in order to not rely on ratechange listener to update the controller.

inject.js
```diff
@@ -484,15 +544,21 @@
         } else if (action === 'faster') {
           // Maximum playback speed in Chrome is set to 16:
           // https://cs.chromium.org/chromium/src/third_party/blink/renderer/core/html/media/html_media_element.cc?gsn=kMinRate&l=166
-          var s = Math.min((v.playbackRate < 0.1 ? 0.0 : v.playbackRate) + value, 16);
-          v.playbackRate = Number(s.toFixed(2));
+          var s = Math.min(
+            (v.playbackRate < 0.1 ? 0.0 : v.playbackRate) + value,
```
ChadBailey: Busted this out to make it clearer what's happening

options.js
```diff
@@ -17,12 +17,15 @@ var tcDefaults = {
     {action: "reset", key: 82, value: 1, force: false, predefined: true}, // R
     {action: "fast", key: 71, value: 1.8, force: false, predefined: true} // G
   ],
-  blacklist: `
+  blacklist:` \
```
ChadBailey: Fixes annoying bug where resetting to defaults always starts with a blank line

options.js
```diff
@@ -163,27 +166,65 @@ function createKeyBindings(item) {
...
   var controllerOpacity = document.getElementById('controllerOpacity').value;
-  var blacklist     = document.getElementById('blacklist').value;
+  var blacklist         = document.getElementById('blacklist').value;
+  var blacklistrc       = document.getElementById('blacklistrc').value;
 
   chrome.storage.sync.remove(["resetSpeed", "speedStep", "fastSpeed", "rewindTime", "advanceTime", "resetKeyCode", "slowerKeyCode", "fasterKeyCode", "rewindKeyCode", "advanceKeyCode", "fastKeyCode"]);
   chrome.storage.sync.set({
-    rememberSpeed:  rememberSpeed,
-    audioBoolean:  audioBoolean,
-    enabled:  enabled,
```
ChadBailey: This section was a confusing mess that seemed to follow no style guide... so i took some style liberties that seemed to make sense in this context.

ChadBailey:
Just wanted to add... This may feel heavy handed to resolve a single site not working correctly. My rationale in making an entirely separate list was because there have been other sites that have been impacted by this as well. I'm thinking of Linked In Learning, but thankfully they honored my request to uncap the playback speed on their end.

igrigorik:
> This adds a second "blacklist", but this list is to block video rate change event listeners on these sites. I have tried it on pluralsight and it works as expected.

Can you elaborate on what the behavior you're after here is?

ChadBailey:
> > This adds a second "blacklist", but this list is to block video rate change event listeners on these sites. I have tried it on pluralsight and it works as expected.
> 
> Can you elaborate on what the behavior you're after here is?

**What**
The intention is to broaden the scope of bugs #521 and #547 to apply to any site the user specifies rather than being explicit to one hard coded domain.

Specifically, it enables playback rate changes on sites that prevent it by applying a different rate change in a function that's listening to the rate change event.

**Rationale/Why**
This presumably impacts only a small subset of sites, pluralsight and linked in learning are the only two I've seen the behavior on personally. That said, it is a big internet and I would imagine it's plausible that many private training/educational sites could also exhibit this behavior.

Importantly I do not think this behavior has been done to intentionally prevent external adjustment of the playback rate, but rather an unintentional byproduct of a design choice by the websites. Since there is no real avenue to provide a native integration, this is the only generic fix I found that worked.

Anyone looking is absolutely welcome to help come up with better ideas about how to handle, it's very possible I've overlooked a better option. I had also considered overloading the playbackrate method, but I'm not certain how to do that in JS or if it's even possible (or if that would be a better approach).

**How it works**
My first thought on how to implement this was to find the specific problematic ratechange listeners and blacklist the ones known to cause issues.

Since [removeEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener) requires the specific function you'd like to remove be supplied to the method, there is no way to remove event listeners without a copy of the original function. I considered trying to iterate over the event listener functions, but also could not find a way to do this. 

This instead puts a dummy listener in place that prevents all event propagation for the ratechange event.

**Caveats/Why use an explicit list**
Since the clickable controller also depends upon a ratechange listener to function, this PR also takes commands that adjust the rate change to directly change the value shown in the controller overlay. This means the controller can become out of sync with the real playback rate if the playback rate has been set by some other means (a native player playback rate adjustment for example). With the implementation used in this PR, the controller being out of sync with the real playback rate should only be possible on blacklisted sites.

Additionally, it's absolutely possible that the site in question has functionality built into the on playback rate event that allows the website to work correctly.

These are the reasons why I consider this a hacky solution that should only be in effect when the user is explicitly requesting it by adding to the list.

igrigorik:
Thanks a lot for the detailed explainer! This is a tricky one.. 

It's unfortunate that their implementation overrides and resets playback rate, but not an unreasonable thing to do either. My intuition is that we should be disabling the controller on such implementations, instead of trying to monkey patch and override their own behavior.. Otherwise, we're entering into a never ending game of unexpected side effects and gotchas. Case in point, having to manually set rate in the controller will cause trouble when you use different built-in controls on various players.

ChadBailey:
> My intuition is that we should be disabling the controller on such implementations, instead of trying to monkey patch and override their own behavior

I'm not sure if this is worth pursuing, but I think we could automatically determine if this is occurring by monitoring the playback rate and seeing if it is changed again within a certain time window. The only problem is, I can't think of a pleasant way to show this is what is happening to the user. Simply removing the controller would cause confusion

> we're entering into a never ending game of unexpected side effects and gotchas.

I definitely understand the concern of beginning a never-ending loop of breaking/patching things. If this feature were enabled and it broke the website's functionality, I wouldn't support further fixes. That's really why I made it a separate list.

Perhaps hiding this list behind "Show experimental features" would be a way to meet in the middle? I would like the ability to _try_ adjusting the playback rate in instances where it's being blocked (presumably unintentionally) by an event listener.

> having to manually set rate in the controller will cause trouble when you use different built-in controls on various players.

I do want to point out that this does not cause problems when you use different built-in controls in players. So long as this "fix" isn't applied, the event listener still updates to show the correct playback rate when the event listener receives a change. I just added a manual update to the listener in order to give an update when the event listener has been disabled. This means a synchronization issue between what's shown in the controller versus actual is only possible when the website has been added to the blacklist.

igrigorik:
I'm glad to hear that LinkedIn updated their implementation. Are you aware of other existing sites where this feature would be useful? 

I could see this being a hidden, advanced / power-user option (that would also require a lot of explanation in the menu) but it would be nice to validate that it works and is useful outside of this one instance. For example, out of the currently open bugs where VSC is not working today, would this address any of the issues?

ChadBailey:
No, I'm unaware of any other sites impacted by this. I decided to implement it in a generic way just to anticipate future need, that wasn't necessarily my expectation. Perhaps it would be best to wait to add it to the interface until we see other examples.

I don't mind updating the PR to adjust accordingly, but don't want to make any changes until there is a clear direction one way or the other. I also wouldn't be mad if this idea was shelved/scrapped as I don't personally use plural sight so it really doesn't impact me. This just happened to be one of the bugs that caught my attention when looking for ways to contribute

igrigorik:
Yep, makes sense, appreciate your great work on this — thanks for digging in!

I'm on the fence. On the one hand, I'd love for the controller to work in as many places as possible, but at the same time this is a special and advanced code path that we'll have to maintain and explain to users. If we had a few sites where this was necessary, I think it would be an easier call. Perhaps one option here is to sit on this for a while.. and see if we get more pluralsight requests, or find other sites where we need the same pattern? Willing to hear arguments in both directions.

wterik:
I am a big pluralsight and linkedin learning user and would very much like this to be integrated.

ChadBailey:
> I am a big pluralsight and linkedin learning user and would very much like this to be integrated.

Thanks for your input, it's helpful to know there are people out there wanting this feature.

> this is a special and advanced code path that we'll have to maintain and explain to users

In my opinion, I don't see any harm in including it under experimental features while disclaiming it as being experimental and advising the user to use at their own risk. I don't see this as being a code path that would need a lot of maintenance or changing, since most of the code changes in this PR are what I would consider optional supporting code. As a reminder the entirety of the functional code in this enhancement is:

```
      document.body.addEventListener("ratechange", function (event) {
      }, true);
```

The rest of the code is just supporting by adding it to the interface and random small improvements I made along the way. I probably should have made those their own PR, sorry for that.

jeffreycli:
Hi ChadBailey, thank you for working on this issue and I came across it when trying to see why this extension wouldn't work on Pluralsight. I've downloaded your fork of this repository and loaded it into my extensions, but Pluralsight is still resetting my playback speed to 1.00x

What steps can I take so that I can watch Pluralsight at the rate that I want?

ChadBailey:
hello, since the discussion has really been around the merits of such a type of fix full testing really hasn't been done. the issue you're describing has been discussed in my above comments though. this fork includes some rework to update the controller with the correct values. that said, the experience will likely never be perfect - which has a lot to do with the quality concerns being brought up.

sorry i haven't had the time to do further investigations on this. i had some unexpected work come up lately so i have had to pause some of the lower priority things like this until those things get resolved.

ChadBailey:
It looks like this may have crept up in a few sites now including twitch.tv (see #584 )

I think I'm going to redo this in a cleaner way and hide behind it behind experimental features as suggested then open as a new PR.

PR #555, opened 2019-12-10
Create Github Action to upload to chrome web store on version tagging

deef0000dragon1:


.github/workflows/chrome-store-upload.yaml
```diff
@@ -0,0 +1,29 @@
+name: Upload To Chrome Web Store
+# This workflow is triggered on pushes to the repository.
+on: 
+  push:
+    tags:
+      - v1*
```
igrigorik: if I'm reading this correctly, this will only trigger for `v1**` pattern-matched tags, correct? Any reason why not to generalize to `v*`?

deef0000dragon1: Not really. A lot of this was back and forth testing of what did and did not work based on examples. Must have just missed it.

.github/workflows/chrome-store-upload.yaml
```diff
@@ -0,0 +1,29 @@
+name: Upload To Chrome Web Store
+# This workflow is triggered on pushes to the repository.
```
jedawson: is this workflow triggered on pushes to the repo or when it detects changes to the version number?

igrigorik: When a tag is pushed. I think you're hinting that we should release the large merge? :)

jedawson:
Maybe I could put it in to a feature flag instead?

PR #556, opened 2019-12-10
Create Github Action to upload dev version to chrome web store on push to master

deefdragon:


igrigorik:
Hmm, may be user error here.. Testing out the publish on tag flow, see results: https://github.com/igrigorik/videospeed/commit/98d938741d158143f3530ba7ae3dceb9cee91d4d/checks?check_suite_id=372107312

Had to update zip params, but looks like it's still including .git files. Also, getting a 404 error for the publish step — broken action?

deef0000dragon1:
The action got updated to V2. though that does not appear to be the issue. I will do some work on my end to see if I can debug it. Best guess is that I messed up the name.

deef0000dragon1:
Also, Do NOT merge this version as it will not function until I add the required manifest editing.

igrigorik:
@deef0000dragon1 couple of things on this one..

* I think I need to create an extension first, and then we can update the name here? 
* Could you update this pull with same params as the release-on-tag PR? 
* We do need a version number when we do releases, we can't just push the same version. How would we handle it here? Auto-increment sub-version number (v.0.5.8.[x])?

deef0000dragon1:
- yes, 
- The idea was for dev extension to be cutting edge/always up to date with master. Requiring a tag might defeat the purpose. 
- that shouldnt be too hard. that or some format based on the unix timestamp,  which i would perfer as it would have some meaning in the context of a development build. Ideally we could just use the commit hash, but from what I looked up, it has to be a.b.c.d (0-65535). I could set version_name to be the hash tho.

igrigorik:
Yep, per https://developer.chrome.com/apps/manifest/version.. 

* Version: one to four dot-separated integers identifying the version of this extension
* Version name: In addition to the version field, which is used for update purposes, version_name can be set to a descriptive version string and will be used for display purposes if present.

We could set version name to the commit hash but version has to be a numeric field. A truncated timestamp is not super helpful imo, so perhaps we could increment the last number?

ChadBailey:
I'll be glad to get this fixed up after my 2 pending PRs #598 and #599 are cleared. I restructured the code in #598 which should make this much easier

ChadBailey:
Thanks so much for doing this and patiently waiting. All pending stuff has been fixed, would love to see this added but reserving next push as a tiny bugfix release. Slated for 0.7 release.

.github/workflows/chrome-store-upload-dev.yaml
```diff
@@ -0,0 +1,31 @@
...
+        uses: montudor/action-zip@v0.1.0
+        with:
+          args: zip -r ./vscd.zip ./
+        
+      - name: Upload to 
+        uses: trmcnvn/chrome-addon@v1
+        with:
+          # extension is only necessary when updating an existing addon,
+          # omitting it will create a new addon
+          extension: video-speed-controller-dev
```
ChadBailey: Sorry I seem to have missed it, where will this be uploaded to/how can we install it or instruct others to install?

.github/workflows/chrome-store-upload-dev.yaml
```diff
@@ -0,0 +1,31 @@
+name: Upload Dev To Store
+
+on: 
+  push:
+    branches:
+      - master
```
ChadBailey: Can we change this to on push to "testing" or "beta" branch? I would like to get in a cadence where we are using a merge strategy that we essentially always have our current stable release, then a beta version we can use for both beta testers who want to help the project as well as to have someone test a fix which we can't test directly. This happens a lot with subscription based video services.

This would also give us an opportunity to try out A|B testing on two styles of implementation

deef0000dragon1:
Actually I Completely forgot about this one. Unfortunately I don't have the brain power or time at the moment to make the changes and debug them. Working from home (guess why, It starts with a C) has basically taken my all. 
Sorry.
Feel free to take over or close.

.github/workflows/chrome-store-upload-dev.yaml
igrigorik: I would need to register a new extension on Chrome Store and then set it up here. The install process would be like any other extension once that's live.

.github/workflows/chrome-store-upload-dev.yaml
igrigorik: I think we can get that with current strategy, no? In effect, we could treat master as tip of tree where each merge triggers a new release for beta channel. That said, one trick here is to figure out auto-increment strategy.

ChadBailey:
> Actually I Completely forgot about this one. Unfortunately I don't have the brain power or time at the moment to make the changes and debug them. Working from home (guess why, It starts with a C) has basically taken my all.
> Sorry.
> Feel free to take over or close.

I understand entirely... I think all of us in the software business are going through that right now. I'll be glad to take this on for you. I appreciate you taking the time and effort to get it going for the project, your efforts definitely paid off.

.github/workflows/chrome-store-upload-dev.yaml
ChadBailey: I guess my thinking was that there may be some bugs we are getting fixed up in the prior release that we would want in a separate build/branch so that a bugfix release could be pushed. If we used the same branch for both it would mean that before tagging and uploading the bugfix release we would have to back out all of the beta features we had in the pipe first

ajayyy:
My beta extension has been getting rejected from the chrome store. Just a warning that Google doesn't like "duplicate" extensions, so you might want to just publish the zip

deef0000dragon1:
I _explicitly_ asked about this like, 2 months ago. its even in their FAQ about duplicate extensions.... 
See [HERE](https://developer.chrome.com/webstore/spam-faq) if you would like to point them to their own policy @ajayyy.

ajayyy:
@deef0000dragon1 Alright, thanks. It might just be because they are doing more automated rejections now due to current circumstances.

PR #563, opened 2019-12-24
Update chrome-store-upload.yaml to use proper name

deef0000dragon1:
Also add exclude to exclude .git files.


deef0000dragon1:
If you want, I can also include the change that I made that edits the manifest to update the version to the tag version.

igrigorik:
Success! 
image


Awesome work, thank you. In theory, v0.5.8 should be staged in Chrome store now.

> I can also include the change that I made that edits the manifest to update the version to the tag version.

That's OK. In practice I have to do some manual updates to the manifest anyway.

PR #583, opened 2020-01-27
Adds adaptive speed functionality

newice:
Press A to toggle adaptive speed.
The speed then increases to double the preferred speed if video is silent and goes back to preferred speed if it is not silent anymore.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -11,6 +11,7 @@
       audioBoolean: false,  // default: false
       startHidden: false,   // default: false
       controllerOpacity: 0.3, // default: 0.3
+      adaptiveSpeed: false,   // default: false
       keyBindings: [],
       blacklist: `
         www.instagram.com
@@ -69,6 +70,14 @@
         force: false,
         predefined: true
       }); // default: G
+      tc.settings.keyBindings.push({
+        action: "adaptive",
+        key: 65,
+        value: 0,
+        force: false,
+        predefined: true
+      }); // default: A
+
       tc.settings.version = "0.5.3";
 
       chrome.storage.sync.set({
@@ -163,17 +172,20 @@
       target.addEventListener('ratechange', this.handleRatechange = function(event) {
         // Ignore ratechange events on unitialized videos.
         // 0 == No information is available about the media resource.
+        // when adaptiveSpeed is activated, there are too many changes.
         if (event.target.readyState > 0) {
           var speed = this.getSpeed();
           this.speedIndicator.textContent = speed;
-          tc.settings.speeds[this.video.currentSrc] = speed;
-          tc.settings.lastSpeed = speed;
-          this.speed = speed;
-          chrome.storage.sync.set({'lastSpeed': speed}, function() {
-            console.log('Speed setting saved: ' + speed);
-          });
-          // show the controller for 1000ms if it's hidden.
-          runAction('blink', document, null, null);
+          if(!tc.settings.adaptiveSpeed) {
+            tc.settings.speeds[this.video.currentSrc] = speed;
+            tc.settings.lastSpeed = speed;
+            this.speed = speed;
+            chrome.storage.sync.set({'lastSpeed': speed}, function() {
+              console.log('Speed setting saved: ' + speed);
+            });
+            // show the controller for 1000ms if it's hidden.
+            runAction('blink', document, null, null);
+          }
         }
       }.bind(this));
 
@@ -241,6 +253,7 @@
             <button data-action="faster">+</button>
             <button data-action="advance" class="rw">»</button>
             <button data-action="display" class="hideButton">x</button>
+            <button data-action="adaptive" class="hideButton">A</button>
           </span>
         </div>
       `;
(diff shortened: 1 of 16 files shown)


inject.js
```diff
@@ -527,11 +540,58 @@
...
         } else if (action === 'jump') {
           jumpToMark(v);
+        } else if (action === 'adaptive') {
+          toggleAdoptiveSpeed(v, controller);
         }
       }
     });
   }
 
+  function toggleAdoptiveSpeed(v, controller) {
```
ChadBailey: I think you may have meant to name this `toggleAdaptiveSpeed`

inject.js
```diff
@@ -527,11 +540,58 @@
...
+        tc.adaptiveSpeedAnalyserNode.fftSize = 512;
+      }
+      const myDataArray = new Float32Array(tc.adaptiveSpeedAnalyserNode.frequencyBinCount);
+    
+      function audioSample(){
+        if(!v.paused){ 
+          tc.adaptiveSpeedAnalyserNode.getFloatFrequencyData(myDataArray);
+          var dbValue = Math.max(...myDataArray) + 100;
+          var quietZone = 45;
+          var eps = 4;
```
ChadBailey: What does `eps` stand for?

inject.js
newice: You are right.

inject.js
newice: I Thought of Machine epsilon. https://en.wikipedia.org/wiki/Machine_epsilon 
But i think it should be named different. I just wanted to have a buffer zone so the speed does not jump around all the time.

ChadBailey:
I tried this out as-is without updating it, and I can't seem to get it to work? Can you provide steps for testing?

ChadBailey:
All of the big changes that required this to be put on hold are behind us now. I've slated for 0.7 release if we can get it worked out before then. @newice  Can you confirm with me how I can test this enhancement?

newice:
How to test the enhancement:
- It does not work on Firefox(https://bugzilla.mozilla.org/show_bug.cgi?id=1517199) , only in chrome
- The key "A" triggers adaptive speed on and off.

Nothing else needed to set up..  @ChadBailey  Did you have any  errors in the console? I also addad a new button to the overlay controls. you can click it instead of the key.

ChadBailey:
> Nothing else needed to set up..

I was able to activate it, but in my tests it didn't actually do anything. I assume maybe it was just the lack of sufficient silence in the videos I tried? I'm unsure. This is why I was hoping for exact steps, like set the settings this way, go to this website, activate, from 0:00-1:50 you should see 1x speed, from 1:50-1:55 you should see 2.5x speed, etc.

newice:
This video is silent between the speaking: https://www.youtube.com/watch?v=kcs82HnguGc
1. Start the Video
2. Hit 'A' ==>When you activate adaptive speed the speed should go from 1 to prefered speed (in my case 1.8)
3. As soon as the speaker stops talking the Speed switches to 2x preferred (in my case 3.6)
4. while active you can not change the speed because it gets resetted by adaptive speed.
5. When you deactivate adaptive speed it jumps to preferred speed (in my case 1.8).


Youtube hides the controls so you maybe don't see the change. Keep the mouse over the video moving to always see the current speed.

ChadBailey:
thanks for the steps for testing. sorry for the delays in getting back to you, i've been busy with COVID-19 related tasks. i plan to test this soon

newice:
Finally i was able to update my pullrequest. So if anyone has the time i would appreciate a feedback.

igrigorik:
@newice this is, fascinating! I took for a quick test drive, a few questions..

* On YouTube my video periodically flashes to 0.0 (full pause) and then resumes -- guessing, that's not intended behavior?
* Have you done any testing on what the CPU impact is when this is enabled?

newice:
> @newice this is, fascinating! I took for a quick test drive, a few questions..
> 
> * On YouTube my video periodically flashes to 0.0 (full pause) and then resumes -- guessing, that's not intended behavior?
This is strange. I never had the speed go to 0.0 For the silent phases i changed the speed from two times the prefered speed to an own setting whichyou can set in the plugin options. Default is 3. Maybe the setting is missing for some reason?
> * Have you done any testing on what the CPU impact is when this is enabled?
No unfortunately i have no experience in extended performance testing. activating the adaptive speed for the first time after page load usually takes about 0.1 second for me.

inject.js
```diff
@@ -436,15 +445,18 @@ function setupListener() {
...
-    tc.settings.speeds[src] = speed;
-    log("Storing lastSpeed in settings for the rememberSpeed feature", 5);
-    tc.settings.lastSpeed = speed;
-    log("Syncing chrome settings for lastSpeed", 5);
-    chrome.storage.sync.set({ lastSpeed: speed }, function () {
-      log("Speed setting saved: " + speed, 5);
-    });
-    // show the controller for 1000ms if it's hidden.
-    runAction("blink", null, null);
+    // when adaptiveSpeed is activated, there are too many changes.
```
igrigorik, line 448: First off, love this! 

Could we make it work such that we don't have to choose between the two? As in, removing silence blocks seems like an additive option in addition to setting a base speed. Some podcasts apps provide this functionality, which you can toggle in addition to speed setting. 

I think we'd just have to distinguish ratechange events that are generated from this logic vs. other sources?

newice: I am totally with you. 
Just made an update. The problem is that ratechange is fired every time.
Now i made an attempt to distinguish before increasing or decreasing speed.

i think there is still a problem when turning off adaptive speed when in silent mode it seems to take the current "adapriveSpeed" as lastSpeed. so video stays at e.g. 5.00 .

newice: i also increased the timeout for the samples to 10ms to hopefully increase the preformance but i randumly seem to get async sound when using it alot.

PR #598, opened 2020-02-28
Added auto test and build CICD pipeline to pull requests

ChadBailey:
Restructured code to separate configs and artifacts from code
Created node project and added testing packages
Added github action for running tests on pull requests
Added demo tests to demonstrate how to use them
Updated CONTRIBUTING.md as needed
Created local pre-commit hook to automatically test and style on push (see screenshot)
image

In the event of a failing scenario (feature), the end user is given a detailed explanation of what went wrong and why like this

image

In the event of a failing unit test, the commit is blocked and the user is given a detailed explanation of the failed test

image

In the event of incorrect styling, the user is notified of the failed style check. At their option, the user can issue the command `npm run fixup` to have their code auto-formatted

image

PR #599, opened 2020-02-28
Fixes #595 by reverting #541 until the issues can be resolved

### LINKED ISSUE #595: Severe performance issues beginning merge #541
I made a faulty assumption that the latest published version of VSC included PR #541 but from what I can see, it's only updated until the commit just before that.

I had planned to but didn't get a chance to test out #541 and now when developing a fix for #584 I was experiencing a severe performance hit. I checked out the last commit before that, and no performance issues up to this commit (I imagine the version bump is what indicated that was the version pushed out to the chrome extension store)

```
commit 0389c61db733e86698edc1e1cd5e4298a2777c1d
Author: Ilya Grigorik <ilya@igvita.com>
Date:   Mon Jan 13 21:45:05 2020 -0800

    bump to 0.5.9
```

I haven't had the opportunity to look at it on other PCs, but this particular laptop I'm on is a pretty powerful gaming laptop so I expect the story to only get worse.

When digging in a little deeper, it seems clear that there is a problem with the recursive call walking the DOM. The PR mentions performance a lot, I imagine this is why the author was concerned. Any change in speed takes roughly 500ms, so rapid subsequent changes results in a compounding effect resulting in the browser being completely locked up for over 200 seconds. Here is what the profile looks like:

image

I haven't had time to look close enough to draw some conclusions yet, but my thought is we should rethink how we approach enumerating controllers (the little clickable box), then how we reach those controllers once the DOM has already been scanned rather than scanning it on every interaction. If this is not a simple fix, we may want to consider also incorporating the architecture concerns I listed in #589 into the fix.

Hoping to take a closer look when I get the time, but wanted to be sure to report this before it gets pushed out to everyone - it will definitely have to be rolled back if others are experiencing this too.

ChadBailey:
Note that this includes #598 so it looks larger than it actually is. You can see [here](https://github.com/igrigorik/videospeed/pull/599/commits/82de6c75415d33c2fb9719790b61170161231a3f) the actual changes which are only to inject.js.


ChadBailey:
Note that this includes #598 so it looks larger than it actually is. You can see [here](https://github.com/igrigorik/videospeed/pull/599/commits/82de6c75415d33c2fb9719790b61170161231a3f) the actual changes which are only to inject.js.

ChadBailey:
I should have based this off of the master branch rather than another pending PR. I'm going to re-open as its own PR

Issue #523, opened 2019-10-01
not working on https://pan.baidu.com


wangbuyao:
im sure the video is HTML5  and this extension runs well on any other web

jacobcolbert:
Is there a way to view a video without an account? Can you link directly to
such a video?

wangbuyao:
https://pan.baidu.com/s/1tfW8rR8RY-nCV75qUKZDZQ&nbsp;
Code：Sze3&nbsp;


thx！


发自我的iPhone


------------------ Original ------------------

jacobcolbert:
Thanks. It works perfectly for me though! Can you open up the console
(Ctrl+shift+J) and let me know if you get any errors?

jacobcolbert:
Can you copy and paste them? I can't figure out what is wrong easily since
on my end it is working fine.

wangbuyao:
I'm sorry for disturbing u.I can hardly understand this puzzle


Service worker active
2userscript.html?id=2b5e4d5f-1e1b-454e-a5a1-637c1ad47fe4:514 waiting
/feproxy/ad/list?list%5B0%5D%5Bid%5D=web-text&list%5B0%5D%5Bw%5D=0&list%5B0%5D%5Bh%5D=0&uk=1655457047&channel=chunlei&web=1&app_id=250528&bdstoken=33cab4479ac5694165b8588905da6c4c&logid=MTU2OTk0Mzk4MDE3MzAuNTA3MjgxNTkwMjI2ODA1NQ==&clienttype=0:1 Failed to load resource: net::ERR_BLOCKED_BY_CLIENT
wappass.baidu.com/v3/login/api/auth?return_type=3&tpl=netdisk&u=https%3A%2F%2Fpcsdata.baidu.com%2Frest%2F2.0%2Fpcs%2Ffile%3Ffrom_module%3Dcloud-ui%26ssnerror%3D0%26errno%3D3%26type%3Dstoken%26traceid%3D%26method%3Dplantcookie%26source%3Dpcsdata%26callid%3D0.1%26errmsg%3Dunknow%2520error%28110001%29%26bduss%3D%26stoken%3D%26logid%3D6354407014121910161&notjump=1&traceid=:1 Failed to load resource: net::ERR_TOO_MANY_REDIRECTS
[...]

PR #604, opened 2020-03-02
Added ratechange cooldown to fix #72 #521 and #584

### LINKED ISSUE #72: Support for Teamtreehouse.com
The speed cannot be adjusted on Teamtreehouse.com The current vido speed shows in VSC but cannot be modified. The speed can only be changed with their built in control which is limited to 2x.

### LINKED ISSUE #521: not working on pluralsight.com

### LINKED ISSUE #584: Twitch live videos not working correctly
Only on live video and not vods the video speed is not able to be adjusted. When pressing the increase speed key you can see it quickly switching to the higher speed but after a few milliseconds it reverts back to default speed(same with ads on a livestream) this wasn't always the case and i think has started in last week or two.

Attached gif is showing the speed identifier where im holding down the increase button and it should be going up by .25

ezgif-3-ca0a00527d3a

ChadBailey:
Based on user feedback and my own testing, I'm confident this will fix #584 which has been reported by many users. This should also fix #72 and fix #521 (NOTE: Originally #523 was claimed to be fixed by this in error, that one appears to be related to the apple.tv shadowroot bug)

~I realize in the past we [went back and forth](https://github.com/igrigorik/videospeed/pull/547#issuecomment-569454746) on implementing this or not, however, with one of the largest video sites on the Internet now being impacted I think it is time to bring this fix in. Implementing it with the warnings clearly shown seems to be the lesser of the two evils.~

~To re-iterate, the only known side effect of this fix is that adjusting using the website's controls (assuming it has any) will not update the controller's speed. Adjusting using VSC will immediately fix this by setting the speed to what the VSC says. This is unavoidable, since all ratechange listeners are blocked including ours.~

Edit: See below for notes on the new technique. The caveats listed above no longer apply.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -374,11 +374,7 @@ function refreshCoolDown() {
   log("End refreshCoolDown", 5);
 }
 
-function initializeWhenReady(document) {
-  log("Begin initializeWhenReady", 5);
-  if (isBlacklisted()) {
-    return;
-  }
+function setupListener() {
   document.body.addEventListener(
     "ratechange",
     function(event) {
@@ -391,6 +387,13 @@ function initializeWhenReady(document) {
     },
     true
   );
+}
+
+function initializeWhenReady(document) {
+  log("Begin initializeWhenReady", 5);
+  if (isBlacklisted()) {
+    return;
+  }
   window.onload = () => {
     initializeNow(window.document);
   };
@@ -448,6 +451,11 @@ function initializeNow(document) {
   // enforce init-once due to redundant callers
   if (!document.body || document.body.classList.contains("vsc-initialized")) {
     return;
+  }
+  try {
+    setupListener();
+  } catch {
+    // no operation
   }
   document.body.classList.add("vsc-initialized");
   log("initializeNow: vsc-initialized added to document body", 5);
--- a/inject.js
+++ b/inject.js
@@ -18,7 +18,12 @@ var tc = {
       vine.co
       imgur.com
       teams.microsoft.com
-      `.replace(regStrip, "")
+    `.replace(regStrip, ""),
+    blacklistrc: `\
+      twitch.tv
+      pluralsight.com
+      teamtreehouse.com
+    `.replace(regStrip, "")
   }
 };
 
@@ -80,7 +85,8 @@ chrome.storage.sync.get(tc.settings, function(storage) {
       startHidden: tc.settings.startHidden,
       enabled: tc.settings.enabled,
       controllerOpacity: tc.settings.controllerOpacity,
-      blacklist: tc.settings.blacklist.replace(regStrip, "")
+      blacklist: tc.settings.blacklist.replace(regStrip, ""),
+      blacklistrc: tc.settings.blacklistrc.replace(regStrip, "")
     });
   }
   tc.settings.lastSpeed = Number(storage.lastSpeed);
@@ -91,6 +97,7 @@ chrome.storage.sync.get(tc.settings, function(storage) {
   tc.settings.startHidden = Boolean(storage.startHidden);
   tc.settings.controllerOpacity = Number(storage.controllerOpacity);
   tc.settings.blacklist = String(storage.blacklist);
+  tc.settings.blacklistrc = String(storage.blacklistrc);
 
   // ensure that there is a "display" binding (for upgrades from versions that had it as a separate binding)
   if (tc.settings.keyBindings.filter(x => x.action == "display").length == 0) {
@@ -306,13 +313,13 @@ function defineVideoController() {
   };
 }
 
-function initializeWhenReady(document) {
-  escapeStringRegExp.matchOperatorsRe = /[|\\{}()[\]^$+*?.]/g;
-  function escapeStringRegExp(str) {
-    return str.replace(escapeStringRegExp.matchOperatorsRe, "\\$&");
-  }
+function escapeStringRegExp(str) {
+  matchOperatorsRe = /[|\\{}()[\]^$+*?.]/g;
+  return str.replace(matchOperatorsRe, "\\$&");
+}
 
-  var blacklisted = false;
+function isBlacklisted() {
+  blacklisted = false;
   tc.settings.blacklist.split("\n").forEach(match => {
     match = match.replace(regStrip, "");
     if (match.length == 0) {
@@ -334,9 +341,46 @@ function initializeWhenReady(document) {
       return;
     }
   });
+  return blacklisted;
+}
 
-  if (blacklisted) return;
+function isRateChangeBlocked() {
+  blockRateChange = false;
+  tc.settings.blacklistrc.split("\n").forEach(match => {
+    match = match.replace(regStrip, "");
+    if (match.length == 0) {
+      return;
+    }
+    if (match.startsWith("/")) {
+      try {
+        var regexp = new RegExp(match);
+      } catch (err) {
+        return;
+      }
+    } else {
+      var regexp = new RegExp(escapeStringRegExp(match));
+    }
+    if (regexp.test(location.href)) {
+      blockRateChange = true;
+      return;
+    }
+  });
+  return blockRateChange;
+}
 
+function initializeWhenReady(document) {
+  if (isBlacklisted()) {
+    return;
+  }
+  if (isRateChangeBlocked()) {
+    document.body.addEventListener(
+      "ratechange",
+      function(event) {
+        event.stopImmediatePropagation();
+      },
+      true
+    );
+  }
   window.onload = () => {
     initializeNow(window.document);
   };
@@ -502,8 +546,10 @@ function initializeNow(document) {
               });
               break;
             case "attributes":
-              if (mutation.target.attributes["aria-hidden"] &&
-                mutation.target.attributes["aria-hidden"].value == "false") {
+              if (
+                mutation.target.attributes["aria-hidden"] &&
+                mutation.target.attributes["aria-hidden"].value == "false"
+              ) {
                 var flattenedNodes = getShadow(document.body);
                 var node = flattenedNodes.filter(x => x.tagName == "VIDEO")[0];
                 if (node) {
@@ -551,6 +597,13 @@ function initializeNow(document) {
   });
 }
 
+function setSpeed(controller, video, speed) {
+  var speedvalue = speed.toFixed(2);
+  video.playbackRate = Number(speedvalue);
+  var speedIndicator = controller.shadowRoot.querySelector("span");
+  speedIndicator.textContent = speedvalue;
+}
+
 function runAction(action, document, value, e) {
   if (tc.settings.audioBoolean) {
     var mediaTags = getShadow(document.body).filter(x => {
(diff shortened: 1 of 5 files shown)


inject.js
```diff
@@ -18,7 +18,12 @@ var tc = {
       vine.co
       imgur.com
       teams.microsoft.com
-      `.replace(regStrip, "")
+    `.replace(regStrip, ""),
+    blacklistrc: `\
```
igrigorik: nitpick: let's use a descriptive name. it's ok to be a bit more verbose if it helps clarity in the code.

blacklistRateChange or something like that.

options.js
```diff
@@ -23,6 +23,11 @@ var tcDefaults = {
     vine.co
     imgur.com
     teams.microsoft.com
+  `.replace(regStrip, ""),
+  blacklistrc: `\
+    twitch.tv
```
igrigorik: Are we sure about this? It seems like they have a good reason to increase rate (to catch up the stream) and we'd break that default behavior. It's not a bug on our part that the video speed jumps to 1.03.. it's exactly what twitch player intends.

inject.js
ChadBailey: I'm usually the one saying this, so I didn't follow my own advice lol. I'll update it.

options.js
ChadBailey: I thought I was sure, but now I'm not so sure anymore...

In my testing I never experienced the playing changing to 1.03x, I only experienced it bouncing back to 1.0x after attempting to change like shown in the gif in #584

ChadBailey: After fairly extensive testing with this fix disabled, I never experienced twitch.tv adjusting the playback rate of the video.

The only time I experienced an adjustment was when the playback buffer went beyond 30seconds, it will occasionally skip to a more recent time.

image

What I suspect may be going on is some kind of adverse effect of using VSC without the fix applied. The original poster did explain the speed would never be fixed from 1.03 causing him to reload the entire page to fix. I would expect if this was intended behavior, it would auto-resolve once the playback has caught back up with the buffer.

ChadBailey: This is odd, it happened quickly when I tested again using the normal production version of the addon found in the Chrome store... and it did adjust back as you mentioned, so it felt like intentional behavior.

image

I'm now left wondering if something else in the updates was impacting this, or if I just got unlucky with my first tests. I'll keep poking at it and see what I can find out.

ChadBailey: Sorry for my monologue, found this one interesting. Either I messed up my tests the first go-around, or what I think is happening is this feature appears to be present on some streams but not others.

Visiting the same stream using the test environment, I was able to immediately see the speed change to 1.03 as it had in the production version of VSC. I didn't make any code or configuration changes between these tests. I suspect we may have stumbled upon some AB testing twitch is doing to see if this feature is worth/ready to push out to the entire user base.

image

Knowing what I know now, I have a few thoughts.

First, we are already interfering with the default twitch playback rate - which may have been why they in turn be why they created a more aggressive mechanism for adjusting speed.

This is done on [line 149 of inject.js](https://github.com/igrigorik/videospeed/blob/master/inject.js#L149)

This has been a part of the behavior [since the initial commit](https://github.com/igrigorik/videospeed/blame/f1827ba63060255ba119431a00d6a26964058f84/inject.js#L61)

Second thought, because the fix does not actually block a ratechange from happening, this fix does not prevent the default behavior of twitch.tv. I validated this by adding a console.log statement before blocking the propagation

```
  if (isRateChangeBlocked()) {
    document.body.addEventListener(
      "ratechange",
      function(event) {
        console.log(
          "Speed changed to: " +
            event.target.playbackRate +
            " (event propogation blocked)"
        );
      },
      true
    );
  }
```

With all of that known, I'm not sure which direction to go. I two valid paths we could take
1. Add twitch.tv to the blacklist as a default, and users are able to move it to the ratechange blocklist if they so desire
2. Add twitch.tv to the ratechange blacklist as a default, and users will be able to use the extension as they expect, without a visual indication of the speed changing to 1.03 then back to 1.0x once the stream is at 1.5 seconds behind the broadcaster.

I suppose the first thing we need to figure out is if we would like to disable the behavior of manually setting the speed to 1.0x every time we initialize videos - which is totally unrelated to the proposed changes here.

igrigorik: > Add twitch.tv to the ratechange blacklist as a default, and users will be able to use the extension as they expect, without a visual indication of the speed changing to 1.03 then back to 1.0x once the stream is at 1.5 seconds behind the broadcaster.

If we implement ratechange blacklist and add twitch by default, my expectation is that the above behavior would not happen: the global listener would catch and block any updates initiated by non VSC logic. Is that not so?

ChadBailey: Good news! I came up with a new method that nullifies the need for a blacklist and should work on any website without interfering!

I think this is a much better fix that addresses your concerns. The new logic works on a timer. When you explicitly change the speed using video speed controller, it will block on rate change events that aren't generated by VSC for a 1 second window. That 1 second window is then refreshed every time a speed adjustment is done in VSC. Once you have stopped adjusting the speed for 1 second, it resumes normal operation.

In my tests, this fully blocks the "bounce back" effect with no ill effect on the regular operation of websites and the fix is seamless.

ChadBailey:
Good news! I came up with a new method that nullifies the need for a blacklist and should work on any website without interfering! I think this is a much better fix that addresses your concerns.

In the new logic, if you make no attempt to adjust the speed using VSC, everything continues working the same way it does today. Once you issue a command to adjust the speed using VSC, this begins a 1-second timer. Rate change events will now be blocked until the 1-second timer completes. That 1-second timer is then refreshed every time a speed adjustment is requested using VSC. Once you have stopped adjusting the speed in VSC for 1 full second, it resumes normal operation.

In my tests, this fully blocks the "bounce back" effect with no ill effect on the regular operation of websites and the fix is seamless.

options.js
ChadBailey: Your expectation is actually not correct because the rate change command does not originate from a rate change listener.

I have come up with an even better solution that nullifies the need for a rate change block list. See my comments below.

inject.js
ChadBailey: This has been removed entirely with the latest version

inject.js
```diff
@@ -551,6 +578,14 @@ function initializeNow(document) {
   });
 }
 
+function setSpeed(controller, video, speed) {
+  var speedvalue = speed.toFixed(2);
+  video.playbackRate = Number(speedvalue);
+  var speedIndicator = controller.shadowRoot.querySelector("span");
+  speedIndicator.textContent = speedvalue;
```
igrigorik: Do we still need these two lines given updated logic?

ChadBailey: Thanks for pointing this out...

Yes, actually this is even more essential now as the ratechange event is blocked, however, this made me realize that now the logic inside of our ratechange event would never be triggered by VSC which I think is the -exact opposite- of the desired behavior.

I have moved this into its own function and now it's triggered irrespective of if the speed change request comes from VSC or if it comes from the site itself. This does bring in to question whether we should be storing lastspeed when it's the website that requested the change rather than VSC, but I've preserved the existing behavior.

If this needs to be changed to only store speeds originating from requests of VSC I certainly don't mind making that change but think it may be better to split into its own bug report since this is how it behaves today.

ChadBailey:
I want to make sure we are on the same page as far as what this change actually does so I've prepared a gherkin formatted feature here. I hope this helps clarify exactly how this (should) behave

```feature
Feature: Bounce Back Protection

    As a user of the video speed controller extension, I want to be able to
    adjust the video playback speed on websites without it being changed back
    immediately so that I can set the speed I prefer

    Scenario: Website sets playback rate on ratechange event
      Given a website twitch.tv that sets the playback rate on change
       When the speed is changed using video speed controller
       Then the ratechange event is triggered
        And the event is blocked
        And the playback rate does not change again

    Scenario: Website explicitly sets playback rate
      Given a website twitch.tv that sets the playback rate on change
        And the user has not used video speed controller to adjust the speed within the last 1000ms
       When the speed is changed and the origin is not video speed controller
       Then the ratechange event is triggered
        And the event is not blocked

    Scenario: Website does not set playback rate on ratechange event
      Given a website youtube.com that does not set the playback rate on change
       When the speed is changed using video speed controller
       Then the ratechange event is triggered
        And the event is blocked
```

ChadBailey:
Don't merge this yet, I'm still testing and it seems to be undoing the fix for youtube changing the playback rate to 1 when you pause/play. I'm guessing/hoping this is an easy fix, but I haven't had time to look closer yet

ChadBailey:
I just spent an inordinate amount of time trying to figure out why YouTube will remember the last playback rate irrespective of the "Remember Playback Speed" setting... only to find out that's YouTube's default behavior lol.

I think this is now ready to be merged, but do feel free to test it out for yourself

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "Video Speed Controller",
   "short_name": "videospeed",
-  "version": "0.6.0",
+  "version": "0.6.1",
```
jedawson: did you do this:
```bash
git tag -a v0.6.1 -m "bump to v0.6.1"
git push origin v0.6.1
```

ChadBailey: I did not... I'll do that

inject.js
```diff
@@ -132,56 +163,54 @@ function defineVideoController() {
...
         } else {
-          tc.settings.speeds[target.currentSrc] = tc.settings.lastSpeed;
-        }
-        target.playbackRate = tc.settings.speeds[target.currentSrc];
-      }.bind(this))
-    );
-
-    target.addEventListener(
-      "ratechange",
-      (this.handleRatechange = function(event) {
```
igrigorik: If we're removing the listener it means our controller will no longer pickup ratechange events from other sources.. e.g. if someone uses built-in YT speed controls, the displayed speed will be out of sync. 

That seems like a regression. Thoughts>

manifest.json
ChadBailey: Tagged, thanks for the tip

inject.js
ChadBailey: I actually didn't remove it, I just combined the two. The existing ratechange listener was placed in the DOM at the location of the video element (this seems wrong to say, but based on my observations this seems to be how an event propagates through the DOM - beginning at the root/document level then flowing down). This was too late, so I had to add another one ahead of it.

I now have combined them, so even if downstream ratechange listeners happen to be blocked by the edge case the controller is updated with the new speed. I will update the gherkin feature to explain the new behavior (the only change is very slight in the edge case)

inject.js
```diff
@@ -334,9 +359,38 @@ function initializeWhenReady(document) {
...
+function initializeWhenReady(document) {
+  log("Begin initializeWhenReady", 5);
+  if (isBlacklisted()) {
+    return;
+  }
+  document.body.addEventListener(
+    "ratechange",
+    function(event) {
+      controller = event.target.parentElement.querySelector(".vsc-controller");
+      rateChanged(controller);
```
ChadBailey: I moved the original logic found in our ratechange into this function `rateChanged`. Since it is no longer being executed as part of controller's object model, I had to refactor it to use the event to find the controller rather than simply being able to use this.video. This allowed me to combine the two.

Originally, the `rateChanged` function was critical because it was manually triggered by speed adjustments via `setSpeed`, however, after combining them this is no longer a critical step.

ChadBailey:
After much testing, I noticed I was getting intermittent errors in the error log pointing to trying to initialize before the document was ready.

I've moved this a little further back in hopes that it will strike a good balance, but since this is by its very nature a race condition I can't be sure this won't penalize us in some situations. It does seem to be working fine on YouTube and Twitch.tv, but I'd like to give it a little more testing time to be sure I don't notice any ill effects. Once I feel confident I'll report back and request a review.

Sorry for so much back and forth, this has been a bit of a complex problem to solve.

ChadBailey:
I have not experienced any ill effects from these changes in my usage. Ready for review

inject.js
```diff
@@ -626,31 +645,9 @@ function setSpeed(controller, video, speed) {
   var speedvalue = speed.toFixed(2);
   video.playbackRate = Number(speedvalue);
   refreshCoolDown();
-  rateChanged(controller);
```
ChadBailey: This call is no longer needed now that the ratechange blocker (to stop "bouncing" playback rate) and ratechange listener (for updating the controller's speed value) have been combined into one. setSpeed inherently calls the logic in rateChanged by virtue of the listener.

ChadBailey:
All issues have been addressed and I've been using this for a while without ill effect. Any objections to merging?

igrigorik:
👍

Issue #616, opened 2020-03-26
Speed indicator not updating in real time


OneMaskedNinja:
When i hold the increase speed button (in my case "+"), it is not updating in real time like it used to. It can take a second or two to catch up.

Additionally this also seems to cause slow down on the entire web page.

Upon further testing refreshing the page seems to fix the issue for that specific instance, but does not prevent the bug from happening again.

ChadBailey:
thanks so much for the report! assigned to our next bugfix release 0.6.3, we'll try to get to it as soon as we can

OneMaskedNinja:
No thank you for the awesome work on this extension!

Issue #589, opened 2020-02-12
Un-clickable in tv.xfinity.com


rsubasic:
Hi.  Thanks so much for a great tool.  Problem is that the controls are hidden behind several layers of stuff on the xfinity player screen.  Possible solution: allow for a new setting to allow for the controls UI to live in different places.  Another issue with xfinity is the transparent layer they put in front.  Perhaps a keystroke to make the control the front layer??

Thanks!

ChadBailey:
## Short Answer
> Target Audience: Non-developers seeking a solid straight forward answer

This appears to be an architecture oversight, so a fix may not happen quickly unless we get lucky and someone has enough time to invest in overhauling the code.

The good news is, this really does need to happen so I don't think we should ignore due to perceived difficulty.

------

## Long Answer

> Target Audience: Developers or non-developers seeking a more in depth understanding of the problem

### Foreword

Hindsight is 20/20... Don't feel like re-writing the below, but noticed I poised this incorrectly. I feel I came across as if I were certain about the issue and solution, but in reality I'm far from an expert in HTML/CSS/JS, so I may be responding out of some misunderstandings or incorrect assumptions.

### Explanation

Unfortunately, this is a byproduct of the combination of 2 architectures. The primary architecture at play is called a [shadow root](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot). The shadow root was introduced to HTML/CSS to provide better encapsulation of the elements so that styling and other types of mutations within that shadowroot don't "bleed" out into the rest of the page.

The reason allowing things to "bleed" out into the other elements is a problem is because inclusion of someone else's code (maybe it was an embedded script, or perhaps you have multiple development teams making different components so they may accidentally call something the same thing). The shadow root was created to solve all of this.

In this case the designers of the website have decided they would like to use shadow roots to structure their website. Now that this has happened, there's a chance that something exists at a higher layer than the shadow root that contains the video. Since the entire point of shadow roots is to not allow things to "bust out" of them, there is no possible way to do this - it's against the intentional design of the browser.

[...]

PR #619, opened 2020-03-28
Manifest Improvements

Gitoffthelawn:
1) Removed deprecated URL.
   a. That old hitbox.tv URL does redirect somewhere else now.  Does that site need to be excluded?
   b. Why are sites excluded here vs. being added to the default exclusions list in Options?
2) Shortened description and added "audio" to reflect new functionality
3) The Firefox fork of this extension benefits from adding:
     "options_ui": {
         "page": "options.html"
     }
Would it hurt to add that to this manifest too?

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -3,7 +3,7 @@
   "short_name": "videospeed",
   "version": "0.6.1",
   "manifest_version": 2,
-  "description": "Speed up, slow down, advance and rewind any HTML5 video with quick shortcuts.",
+  "description": "Speed up, slow down, advance and rewind HTML5 audio/video with shortcuts",
   "homepage_url": "https://github.com/igrigorik/videospeed",
   "icons": {
     "16": "icons/icon16.png",
@@ -28,8 +28,7 @@
       "exclude_matches": [
         "https://plus.google.com/hangouts/*",
         "https://hangouts.google.com/*",
-        "https://meet.google.com/*",
-        "http://www.hitbox.tv/*"
+        "https://meet.google.com/*"
       ],
       "css": ["inject.css"],
       "js": ["inject.js"]


ChadBailey:
> 1. Removed deprecated URL.
>    a. That old hitbox.tv URL does redirect somewhere else now.  Does that site need to be excluded?
>    b. Why are sites excluded here vs. being added to the default exclusions list in Options?

I don't actually know, and have wondered this myself. Since I haven't been around since the beginning maybe @igrigorik can shed some light on why?

> 2. Shortened description and added "audio" to reflect new functionality

I like these changes

> 3. The Firefox fork of this extension benefits from adding:
>    "options_ui": {
>    "page": "options.html"
>    }
>    Would it hurt to add that to this manifest too?

I'm apparently out of the loop on this one, what will this do?

As an aside, I also didn't realize there was a firefox fork. Maybe you or someone else could explain why we couldn't update this codebase to support both Chrome and Firefox? I can't speak for @igrigorik but I think it would be great to do and those supporting the Firefox fork could contribute here? *shrugs*

codebicycle:
At version 0.5.9 the Chrome and Firefox codebases were closer than ever.
Only 3 changes were needed to port version 0.5.9 to Firefox. Two in the manifest and another one in the code itself. All changes are compatible with Chrome.

Unfortunately, version 0.6.1 brings an unacceptable performance hit in Firefox. Controller lags when changing speeds. Solutions need to be found here.

The `options_page` key from the manifest is deprecated in favor of `options_ui`
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/options_page.

Here's the relevant portion from the Firefox VSC manifest file:

    "options_ui": {
      "page": "options.html",
      "open_in_tab": true
    }

igrigorik:
Don't have any objections to adding the FF option as long as it doesn't cause issues with Chrome. Curious to know what the performance issues are.. I'd suggest we merge this as is and continue other parts in separate thread.

ChadBailey:
> Curious to know what the performance issues are.. I'd suggest we merge this as is and continue other parts in separate thread.

I'm quite sure they are related to #595 as the fix never came close to offering the original performance. Others are reporting as well, as in #616 . A few things to keep in mind is there is a big difference in computer speeds and user expectations.

I had already planned to change it so that setspeed actually directly changes the controller rather than waiting for the ratechange event to fire which should make it feel a bit more snappy, but even when it was that way, the getshadow/getchild logic is just very slow in comparison to previous performance. The problem I have is, I don't actually know enough about JS to debug this so unfortunately, you all are on your own on this one. I tried really had to figure it out when I posted #595 but could not. @jedawson may be able to shed some more light on this. Here is a snapshot of a performance profile that shows how many times the function gets called on YT - it seems like it is being called far more often than necessary, but I really have no idea.

image

A part of the restructuring effort I mentioned in #589 was aimed at solving this. I think the root cause is the need to do an exhaustive DOM search for every single call to getshadow/getcontroller. I think @jedawson may have mentioned this somewhere in his PR as well. When we get the reference, we should be somehow storing it as part of the object model to be reused (if I'm understanding this right - again this is honestly beyond my current understanding of JS).

Any help cracking this one would be deeply appreciated.

ChadBailey:
> I'd suggest we merge this as is and continue other parts in separate thread.

Agreed, going to land this one since it is straight forward. I'll try and remember to open the other things as their own issues... unless you want to, I would be grateful.

PR #621, opened 2020-03-28
Improvements to Options page

Gitoffthelawn:
Thank you for the great extension.  Here are some self-explanatory improvements to the Options page.

### DIFF
--- a/options.html
+++ b/options.html
@@ -25,7 +25,7 @@
         <input class="customValue" type="text" placeholder="value (0.10)" />
         <select class="customForce">
           <option value="false">Do not disable website key bindings</option>
-          <option value="true">Disable websites key bindings</option>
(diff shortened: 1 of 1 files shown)


ChadBailey:
These grammatical issues have bothered me a while, but not enough to actually fix haha. Thanks for doing this!

I've added this to 0.6.3 and I'll take a closer look before merging, but don't see anything that needs to be changed.

ChadBailey:
Looks like there may be a couple of other places that could use some fixing up if you want to add them to this

i.e. https://github.com/igrigorik/videospeed/blob/c6fbdc70d4f2c736deea1c47c51535904fd2f67a/options.js#L151

ChadBailey:
Did not notice the change I pointed out was in the other PR. Good work, thank you!

Issue #609, opened 2020-03-08
Controller disappears when some mutations occur


ChadBailey:
Formerly: Old controller isn't fully removed (renamed title for clarity)

Steps to reproduce:
1. Install latest version of VSC master branch (last commit is [5a53f81](https://github.com/igrigorik/videospeed/commit/5a53f81cba0091a343ddcaca8a9bad8853c65eb7)) - note that this isn't in the store yet for some reason
2. Visit https://app.pluralsight.com/course-player?clipId=1e8b695e-a148-4b6b-83d0-07c9ddbb9b73
3. Click the settings gear on the bottom-right (speed control box works too, anything that triggers a mutation)
4. This will trigger `MutationObserver` and when it reaches `oldController.remove();` the controller is removed, but only partially.

This line of code should have been executed but doesn't seem to be:
https://github.com/igrigorik/videospeed/blob/5a53f81cba0091a343ddcaca8a9bad8853c65eb7/inject.js#L220

Also, no errors to speak of, and breakpoints for some reason aren't triggered, but the controller does in fact get removed.

5. `checkForVideo` is then executed which in turn executes `node.vsc = new tc.videoController(node, parent);`
https://github.com/igrigorik/videospeed/blob/5a53f81cba0091a343ddcaca8a9bad8853c65eb7/inject.js#L516
6. Finally, on line 125 it checks for "vscid" which it finds because `oldController.remove();` didn't complete all of its instructions on step 4 and the new video controller is never generated:
https://github.com/igrigorik/videospeed/blob/5a53f81cba0091a343ddcaca8a9bad8853c65eb7/inject.js#L125

Workaround: comment out `oldController.remove();` on line 514:
https://github.com/igrigorik/videospeed/blob/5a53f81cba0091a343ddcaca8a9bad8853c65eb7/inject.js#L514

cristiancundari:
This happens because of this check on videoController object creation:

https://github.com/igrigorik/videospeed/blob/884be009a45b80a2ea95ac28391c4e2b53e220bd/inject.js#L156-L158

When the oldController get removed, the vscid data attribute on the video node will not.

It can be solved by deleting the previous instance of vsc from DOM video element; simply by adding this code

```javascript
if (node.vsc) {
  delete node.dataset.vscid;  
  delete node.vsc;  
}
```

after `oldController.remove();` on line 602:

[...]

Issue #622, opened 2020-03-28
bbc player


marygab:
It doesn't work with bbc player and it used to

ChadBailey:
Thanks for your report! Can you link an example video with instructions to reproduce? I don't use it and would like to be sure I'm able to test it.

I think it could be caused by #609 but need to confirm

Issue #618, opened 2020-03-28
Facebook Videos No Longer Works


DanitiKain:
Looks like Facebook no longer works with this tool. It used to work.

ChadBailey:
Thanks for the report. I don't know but think this may be related to #609 which has an open PR to fix. I'll test it myself and make sure my hypothesis is correct

ChadBailey:
I'm unable to reproduce this, Facebook is working for me. Can you provide steps to reproduce?

Here are the steps I took:

1. Visit https://www.facebook.com/watch/
2. Click play on whatever video shows up
3. Use hotkeys and controller to adjust speed, both work

DanitiKain:
It appears that it still works on FB classic mode. However, if you switch to their new interface it no longer works. Click top right to switch modes. There is an option to switch to new facebook.

ChadBailey:
I can't find the option to switch modes, perhaps the new mode is just being rolled out to limited sized trial groups?

DanitiKain:
That must be the case. https://www.androidpolice.com/2020/03/19/facebooks-new-web-interface-is-rolling-out-for-some-with-dark-mode-in-tow/

halinagold:
Just to confirm that I see the issue as well, with the new Facebook design (which I'm not able to rollback), on my Windows 10 laptop. Until then this extension worked like a dream, so helpful. Thanks!

PR #623, opened 2020-03-28
Fixing #609

cristiancundari:
Clear the previous instance of vsc from DOM video element when videoController gets removed.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -600,6 +600,10 @@ function initializeNow(document) {
                   )[0];
                   if (oldController) {
                     oldController.remove();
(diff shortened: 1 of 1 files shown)


ChadBailey:
Thanks so much for doing this! I'm guessing it works exactly as advertised, but going to test it then get back

cristiancundari:
Sure, I've tested it and seams to work properly.

ChadBailey:
Thanks for the fix! I've tested it and it looks good. I'm going to do this as a hotfix, because I think the other 2 linked issues #622 and #618 may be related.

PR #628, opened 2020-03-30
Replace deprecated `options_page` key from the manifest

codebicycle:
The `options_page` key from the manifest is deprecated in favor of `options_ui`
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/options_page.

This is from the Firefox port of VSC and shouldn't afect Chrome in any way.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -11,7 +11,10 @@
     "128": "icons/icon128.png"
   },
   "permissions": ["activeTab", "storage"],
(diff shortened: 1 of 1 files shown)


ChadBailey:
Thanks so much for your PR, taking a look now!

ChadBailey:
Looks good, tests good, simple enough I think we can safely move forward with this

Issue #626, opened 2020-03-29
Kaltura issues 


belvedere3:
with kaltura video player, none of the buttons are working now

ChadBailey:
i've never heard of kaltura, but searched for it and found https://corp.kaltura.com/. the videos there work for me, can you provide steps to reproduce?

MajinBui:
In my case, Kaltura is used internally by Georgia Tech for serving lecture videos.  A public Kaltura video can be found [here on Kaltura's website](https://videos.kaltura.com/media/SAP+-+Using+Video+Across+the+Entire+Organization/1_1yzzxwji).

There are two issues that Kaltura's players have, one that I have solved and another that needs more discussion with a partial solution:
- The overlay on Kaltura videos does not update when the speed is changed through their own speed controller and through VideoSpeed's controller
   - VideoSpeed does actually increase the speed, but doesn't show it visually since the setSpeed function does not update the overlay and only the rateChange event
- Kaltura videos set the video speed values to their own controller's values every time a media event is triggered (play, seek, pause) and do not emit a ratechange event.  This causes the overlay to incorrectly display an incorrect value.
   - A special behaviour is found with Kaltura seek event.  Compared to normal Youtube videos and the like, Kalura only emits a seeked event, and not a play event (event though the video is played).


For the overlay not representing the correct speed values, I have a PR ready for this.  For the issue where Kaltura emits their controller's speed value on every media event, there isn't an easy way around it.  Would a good idea be to introduce a new settings flag for VideoSpeed to force apply their speeds instead of the video controller's?

PR #654, opened 2020-05-13
Affects #626 fixing Kaltura video players

MajinBui:
### What has been done?
- Added an event listener for seeked video events and mirrors the
actions of a play event
- Overlay will now update when keypress speed changes are detected
- When the VideoSpeed controller isn't found, the extension will now
look for it in the video element as well as fallback

### Why has this been done?
- Kaltura videos act differently in a few cases.  On seek events, a play
event is not emiited. On speed change events, a rateChange event is not
emitted.  Because of this, it causes the overlay not to update on
keypress or seek events.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -183,35 +183,47 @@ function defineVideoController() {
 
     this.div = this.initializeControls();
 
+    var mediaEventAction = function(event) {
+      storedSpeed = tc.settings.speeds[event.target.currentSrc];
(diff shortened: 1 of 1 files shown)


ChadBailey:
Thanks so much for your contribution! So sorry for not responding sooner... All tests well, so I think this is ready to go in to our bugfix release

PR #669, opened 2020-06-04
Fixes #666, speed lost after pause

tarunlalwani:
Fixes #666, speed lost after pause

PR #678, opened 2020-06-25
fix for tofixed on string error

### LINKED ISSUE #666: speed resets if we pause the video
I just increased my speed to 2.5 and paused the video for a while and then played it again. The speed has been reset to 1.0 even after toggling remember playback speed.

pepijnmm:
Fixes #666,
Problem: tc.settings.speeds[src] became a string which toFixed  can't handle.
Now before getting saved in tc.settings.speed[src] it gets converted to a number.

Line 419 is added so you see 2.00 instead of 2 in the top corner.

Made a merge request because tarunlalwani didn't respond anymore on his pull request.(made a comment 20 days ago) https://github.com/igrigorik/videospeed/pull/669

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -411,12 +411,12 @@ function setupListener() {
       var video = event.target;
       var speedIndicator = video.vsc.speedIndicator;
       var src = video.currentSrc;
-      var speed = video.playbackRate.toFixed(2);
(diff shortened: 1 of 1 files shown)


pepijnmm:
When does the chrome extension normally get updated(because people are still making issues about it)?
It was updated one day before this merge.

PR #689, opened 2020-07-20
proof-of-concept: display saved duration in extension icon's badge

siadat:
As promised in #185, here is a proof-of-concept for showing the total saved time on top of the extension icon. I wanted to see how it could work. My conclusion is that it adds too much complexity for a feature that is not of critical importance for the maintainers of the project. So, I don't recommend it being merged. However, feel free to use it as the basis for your own experimentations. :)

Here's a screenshot of the badge indicating 9 minutes of saved time (in total, for all videos in all tabs since the extension was installed).

Screen Shot 2020-07-21 at 00 43 14

### DIFF
--- a/background.js
+++ b/background.js
@@ -1,0 +1,46 @@
+let savedTimerFormatDuration = function(ms) {
+  let seconds = ms / 1000;
+  let h = Math.floor(seconds/3600);
+  let m = Math.floor((seconds - 3600*h) / 60);
+  let s = Math.floor(seconds - 3600*h - 60*m);
+  console.info(`saved duration: ${h}h:${m}m:${s}s`);
+  if (h === 0 && m === 0) {
+    return `${s}s`;
+  } else {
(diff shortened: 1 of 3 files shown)


igrigorik:
Thanks for prototyping this! Two thoughts.. 

There's been some suggestions around displaying time remaining, given the current speed, instead of time saved -- I think the former makes a lot more sense, as saved is just a vanity metric but remaining provides useful information. That said, if we're going to use the badge space, I think my preference is to reserve it for "current speed" -- shifting this display into the UI would fix a lot of potential problems for folks.

With above in mind, I think my suggestion is to park this but we can use it as reference in the future.

siadat:
> displaying time remaining, given the current speed
That sounds good, I would love that too. It would probably be simpler to implement as well (fewer states/information to handle).

Feel free to close/open this thread any time you want :)

Issue #690, opened 2020-08-02
Allow configurable controller z-stacking to bypass overlay objects


hakt0-r:
Sites like...
* primevideo.com
* iview.abc.net.au 
* sbs.com.au

have multiple layers separating their video from their overlay controls or other objects. This renders the use of mouse with VSC useless. 

Solution: 
A simple fix is to use the switch case statement and start adding exceptions. The **_smarter_** way would be to add a setting where we can play with the **z-Order** of the VSC-controller div tag, like a _bring-to-forward_ **_or_** _send-to-back_ one step/layer at a time, per site that is having issues and it then somehow remembers that setting. I am not _that_ skilled in JS sadly to implement the advanced solution, but I have implemented the easier version, at least for me.

pepijnmm:
same problem with facebook: #693

igrigorik:
> The **_smarter_** way would be to add a setting where we can play with the **z-Order** of the VSC-controller div tag, like a _bring-to-forward_ **_or_** _send-to-back_ one step/layer at a time, per site that is having issues and it then somehow remembers that setting.

I think this is the right way to solve this problem. The list of exceptions continues to grow and it is brittle and unmaintainable. I'll open a new issue for us to track this.

pepijnmm:
I would say that implementing both options are good.
If I'm a new user it would be nice that websites that get used a lot like: youtube, facebook, netflix etc already work without setting up. Of course the question is when will a website be popular enough to get added to the list but a lot of users that use this extension don't know how to code so they won't make a pull request for a website that they use a lot.

There will then of course still be a lot of websites that won't work because of overlays, than the keyboard back and forth will be nice. Then when Facebook will update their website again videospeed users could still use the website with using the back and forth to fix it so they don't have to wait till the extension gets updated.

But that is just my opinion I understand that saving and keeping it updated is not something that you would want but in the last year I haven't seen a lot of changes made to fix the controller on overlays.

igrigorik:
Yep, that's reasonable. We can commit the defaults for small set large and popular sites, but defer the rest to the user.

PR #691, opened 2020-08-02
fixes issue #690: site specific vsc-controller stacking.

hakt0-r:
The proposed changes should address the overlay issue on the few sites I've mentioned on the issue #690

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -342,6 +342,8 @@ function defineVideoController() {
 
     switch (true) {
       case location.hostname == "www.amazon.com":
+      //fixes issue #690: site specific vsc-controller stacking.
+      case location.hostname == "iview.abc.net.au":
       case location.hostname == "www.reddit.com":
       case /hbogo\./.test(location.hostname):
         // insert before parent to bypass overlay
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -347,6 +353,13 @@ function defineVideoController() {
         // insert after parent for correct stacking context
         this.parent.getRootNode().querySelector(".scrim").prepend(fragment);
 
+      //fixes issue #690: site specific vsc-controller stacking.
+      case location.hostname == "www.primevideo.com":
+      case location.hostname == "www.sbs.com.au":
+        // insert before parent to bypass overlay
+        //log("Adding exceptions for ggParent layer stacking.", 5);
+        this.parent.parentElement.parentElement.insertBefore(fragment, this.parent.parentElement.parentElement.firstChild)
+
```
pepijnmm, line 362: should there not be a break; here?(only not sure if tv.apple will then still work because it also doesn't have a break;)

igrigorik:
Per discussion in #690, let's park this in favor of customizable z-stacking.

PR #702, opened 2020-09-08
Design update

nickdvlpr:
Hey there!

This extension has saved me days worth of my time, so I wanted to give back and share a few small edits I made for my own personal use in case others wanted something similar. **This fork changes the style to look more like Google's Material Design.**

I left it as a pull request in case you wanted to implement it, otherwise feel free to close it! Thank you for making this extension; it has significantly improved my life and the lives of my classmates.

image
image

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
-  "name": "Video Speed Controller Redesign",
+  "name": "Video Speed Controller",
   "short_name": "videospeed",
-  "version": "0.1",
+  "version": "0.6.3",
   "manifest_version": 2,
   "description": "Speed up, slow down, advance and rewind HTML5 audio/video with shortcuts",
   "homepage_url": "https://github.com/igrigorik/videospeed",
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
-  "name": "Video Speed Controller",
+  "name": "Video Speed Controller Redesign",
   "short_name": "videospeed",
-  "version": "0.6.3",
+  "version": "0.1",
   "manifest_version": 2,
   "description": "Speed up, slow down, advance and rewind HTML5 audio/video with shortcuts",
   "homepage_url": "https://github.com/igrigorik/videospeed",
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
-  "name": "Video Speed Controller",
+  "name": "Video Speed Controller Redesign",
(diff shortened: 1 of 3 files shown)


igrigorik:
❤️ 🤟 ❤️ ... awesome work, thank you sir! love it.

nickdvlpr:
Thanks! I'm glad you like it. The TODO is something I haven't figured out how to animate cleanly yet and was more of a reminder to myself to work on it later.

Sidenote: One potential drawback of this style is that it lacks any contrast on white backgrounds since it is white text on a white semitransparent background.

igrigorik:
Oh hmm, that's a good point. That said, same problem applies for default YT controls -- right? I don't think they do anything special to auto-detect and invert the color scheme.

nickdvlpr:
You're right they don't appear to use anything special although there is a slight black gradient at the bottom of the video when you hover over it. A similar effect could maybe be achieved by adding a drop shadow behind the controller. I'll play around with it when I get the chance in a few days!

igrigorik:
Took out the committed version for a spin and, first got tripped up by opacity. My default was set to 0.3 in my settings and that made it basically impossible to see with the new styling -- an indicator of difference in contrast. Once I flipped it to 1.0 it definitely got better, but it is still hard to spot on some videos. On dark backgrounds, it looks great. On light backgrounds, it's hard to see. 

First reaction to the solid white mock: it would certainly (I think) fix the contrast issue but I'm a bit worried about the controller becoming an attention target -- occluding the video, etc. At a minimum, colors aside, I do really like the changes you made to how the controller is displayed. Even if we pass on changing the colors, I think those are still worth landing.

Hmm...

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "Video Speed Controller",
+  "name": "Video Speed Controller Redesign",
```
pepijnmm, line 2: The name doesn't have to change and the version can be changed to 6.4 not 2.0

nickdvlpr: Ah, I apologize I didn't intend to commit that change to this pull request. I was in the process of replacing this repository with an update based on @igrigorik 's last reply when you commented. Please see the new Design Update pull request here: #705

nickdvlpr:
@igrigorik I agree, while I prefer this style, it does have its drawbacks. I created a new Pull Request without these drawbacks that has a style in between the current version (0.6.3) and my attempt here.

Please see the new Design Update pull request here: #705

PR #708, opened 2020-09-21
Speed improvements + code cleanup

asselin:
- Remove getController and the overhead associated with it
- Document the data structures
- Remove the vscid property, since it's no longer needed


asselin:
@igrigorik This is the updated PR for the speed improvements. Have you had a chance to look it over?

igrigorik (commented):
Overall, this looks great! 

One caveat is that I've not tested this out in the wild. Unfortunately, looks like there is another merge conflict due to a PR we merged last week — doh! Any chance you could rebase against latest?

Separately, do you have any particular examples or benchmarks where you were testing this code? What did you observe in terms of difference in overhead or overall performance?

asselin:
@igrigorik Merged.

It was about 6 months ago, but what I did was use chrome devtool's performance tab to determine how much CPU time it took to respond to a keypress Event on a site with and without a video.  In both cases, since we don't have to enumerate every element of the DOM anymore, it was like 100x faster.  If you want it for the release notes, I can do a before/after comparison.

IIRC (and again, it was 6 months ago), I was looking into that because there was a note about this extension being forked because of keypress events being unbearably slow on firefox.

igrigorik (approved):
Thanks for the patience and great work here!

igrigorik:
BTW, thanks that all makes sense. Merged and will test drive it locally for a while, will report back if I hit any issues.

pepijnmm:
the line 733:     var controller = v.vsc.div;
error: inject.js:733 Uncaught TypeError: Cannot read property 'div' of undefined
after clicking next video button on pluralsight the vsc can't be find in the video object.
example: https://app.pluralsight.com/course-player?clipId=d889e0b6-67ec-41c1-b47f-e84c927367b0
if you are done with watching Introducing c# and .net - summary you get a button "Continue to next module" if you click on that you will go to learning the c# syntax but you will notice that you can't change the speed anymore.
Also got the error on line 715 "  var speedIndicator = video.vsc.speedIndicator;"
when changing the position of the video after going to the learning the c# syntax part.

A sollution would maybe be the getController again but haven't tested it only reverted to #710 to test if pluralsight worked there(it did).

asselin:
@pepijnmm I fixed a similar kind of issue with this commit: https://github.com/asselin/videospeed/commit/cb30d4a8f163b433525541a39ce1b6a29a05fdf5.  Given what you're describing, I wonder if there's a wider issue that is timing related (e.g. a newly created video element gets a speed change event before we create a controller for it).  

Thank you for reporting the issue with steps to reproduce; I'll take a look at it!

cc: @igrigorik

pepijnmm:
> @pepijnmm I fixed a similar kind of issue with this commit: [asselin@cb30d4a](https://github.com/asselin/videospeed/commit/cb30d4a8f163b433525541a39ce1b6a29a05fdf5). Given what you're describing, I wonder if there's a wider issue that is timing related (e.g. a newly created video element gets a speed change event before we create a controller for it).
> 
> Thank you for reporting the issue with steps to reproduce; I'll take a look at it!
> 
> cc: @igrigorik

Have you found time to take a look? Sorry for the bump 😅

asselin:
@pepijnmm I'll take a look tonight

asselin:
Quick update on this @pepijnmm .  It doesn't look like my changes broke pluralsite-- if I run the version from Sep 18 (4eee3d6851bbebba8891ca707481536d1ca98f4c), it's broken.  I'll see if I can figure out what's going on.

pepijnmm:
> It doesn't look like my changes broke pluralsite--
Pluralsight already gave errors before your commit and had problems with resetting the speed when changing it.
At commit: https://github.com/igrigorik/videospeed/commit/ddae3e093513a8e58626dca5ce4406f03388590e (on commit before yours) I could still change the speed after pressing "Continue to next module". which didn't work anymore after your commit. All the other problems with Pluralsight were already there before your commit, sorry if I insinuated anything else, I didn't mean to.(like resetting the speed when changing)
#730 Seems to fix it, thx!

PR #710, opened 2020-09-25
Fixes iframe flakiness on Funimation

captainsaltcracking:
Often, when a video is in an iframe, the extension behaves strangely. The controller doesn't appear, and the speed isn't remembered when the video is paused and unpaused. This change fixes both of these issues. Also, Prettier made a few small formatting changes.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -235,7 +235,7 @@ function defineVideoController() {
           (mutation.attributeName === "src" ||
             mutation.attributeName === "currentSrc")
         ) {
-          var controller = getController(this.id);
+          var controller = getController(this.id, target);
           if (!controller) {
             return;
           }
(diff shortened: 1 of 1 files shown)


igrigorik:
Thanks for the contribution! Overall, it looks good. Do you have an example URL where we can test the before/after?

captainsaltcracking:
Sure! You can see it on Funimation videos like this one:

https://www.funimation.com/shows/sword-art-online/the-world-of-swords/

It's non-deterministic, but you should be able to see the failure within 10 refreshes or so.

igrigorik:
Nice, thank you! I can definitely see the before and after there. ty again for the great contribution!

igrigorik (approved):

PR #730, opened 2020-11-21
Fix deletion when aria-hidden is set

asselin:
This fixes the error issue that @pepijnmm found (described in #708 ).  It does not fix the issue with pluralsite resetting the speed.  The best I can tell, I think the pluralsite player resets the speed to whatever it's control is set to after videospeed sets it.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -652,15 +652,8 @@ function initializeNow(document) {
                   (x) => x.tagName == "VIDEO"
                 )[0];
                 if (node) {
-                  var oldController = flattenedNodes.filter((x) =>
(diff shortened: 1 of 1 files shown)


pepijnmm (approved):
tested on youtube, facebook, netflix, pluralsite.
Pluralsight problem as Asselin mentioned seems fixed(not related to speed change) and all the other sites work just like before so that is good.

igrigorik (approved):
👍

igrigorik:
Looks great, thanks for tracking it down!

PR #752, opened 2021-01-21
fix control position on hulu w/ getBoundingClientRect 

### LINKED ISSUE #562: Hulu browser reload causes controller to move to center
Occasionally, the controller will show up in the center of the screen. I've had it happen both on Chrome and on Brave.

This appears to happens everytime, when I reload a Hulu tab in the browser that has a video with the controls showing up. Killing the tab and reloading Hulu, or restarting the browser, fixes the issue. However, closing all the video overlays (including the smaller overlay the video reduces to), so Hulu is back to the browsing page, then reopening a new video, does NOT fix the issue.

Browsing from that tab to another streaming site DOES reset the controller position to default. I've tried with Netflix and Amazon Prime.

May be related to #536

### LINKED ISSUE #744: Hulu speed menu stuck in middle on Hulu.
On hulu the speed menu is stuck in the middle, you can drag the speed menu somewhere else or you can hide it, but as soon as you move your cursor and the pause menu pops up, the speed menu goes back to the middle.

paulirish:
On hulu's desktop web client, the video is stretched to "fullscreen" with transforms!

image

This means the technique to correctly position the controls puts them right in the middle of the video.

`getBoundingClientRect` takes transforms into account, but otherwise I think it has consistent semantics with `offsetLeft`/`offsetTop`. (i tested with adding extra margin and extra border and box-sizing..... and the results are the same)

I tested with hulu, youtube, netflix, facebook and they're all lookin fine.

fixes #562. fixes #744

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -272,8 +272,8 @@ function defineVideoController() {
   tc.videoController.prototype.initializeControls = function () {
     log("initializeControls Begin", 5);
     const document = this.video.ownerDocument;
-    const rect = this.video.getBoundingClientRect();
     const speed = this.video.playbackRate.toFixed(2);
+    const rect = this.video.getBoundingClientRect();
     const top = Math.max(rect.top, 0) + "px";
     const left = Math.max(rect.left, 0) + "px";
 
--- a/inject.js
+++ b/inject.js
@@ -271,10 +271,11 @@ function defineVideoController() {
 
   tc.videoController.prototype.initializeControls = function () {
     log("initializeControls Begin", 5);
-    var document = this.video.ownerDocument;
-    var speed = this.video.playbackRate.toFixed(2),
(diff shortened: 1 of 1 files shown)


igrigorik (approved):
woohoo, glad this is finally fixed — ty!

pepijnmm:
@paulirish For me this merge breaks the youtube support controller position(just noticed hadn't updated in a while)
I tested compared to enforce normal white-space rendering for controller (#753) so only this commit changed things
Before merge:
Screenshot_8
after merge
Screenshot_7

paulirish:
> For me this merge breaks the youtube controller position(just noticed hadn't updated in a while)

yeah i can reproduce this. you're right.  :)
thanks for isolating it. 

yeah this is going to require some more thinking.. finding a positioning calculation that's respectful of offsetParent but still gives results.

PR #753, opened 2021-01-21
enforce normal white-space rendering for controller

### LINKED ISSUE #746: Wrong controller UI shape in embedded videos on Discord
As you can see, the speed controller UI is misshaped and clips off the thumbnail. Also I think it should not show up until the video is actually playing.
speed error

paulirish:
fixes #746 

| before | after |
|-|-|
| image |image |

note on the css placement: while there are origin-specific fixes, it feels like this defensive style can afford to be _always_ be present. 
(also, placing the style within shadow.css was my preference, unfortunately it's incomplete: while applying to #controller fixes the layout of the buttons, there's still excessive whitespace remaining)

### DIFF
--- a/inject.css
+++ b/inject.css
@@ -9,6 +9,11 @@
   opacity: 1 !important;
 }
 
+.vsc-controller {
+  /* In case of pages using `white-space: pre-line` (eg Discord), don't render vsc's whitespace */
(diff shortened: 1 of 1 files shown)


igrigorik (approved):
👍 thanks Paul!

Issue #754, opened 2021-01-24
Videospeed enlarges the videos at honeycode.aws


russelldavis:
Link: https://www.honeycode.aws/

This messes up the site's layout and adds horizontal scrollbars.

Screenshot: 
<img width="1329" alt="Screen Shot 2021-01-24 at 2 28 57 PM" src="https://user-images.githubusercontent.com/551404/105645866-44871f00-5e51-11eb-9de4-d336c9129dec.png">

pepijnmm:
I know what the problem is but not sure how to fix it on our end. they use css that applies on only-child on the video element. but because we added our vsc-controller to it it doesn't apply the css anymore on the video element
the css that doesn't work anymore after we vsc-controller is added
![image](https://user-images.githubusercontent.com/3217184/105647056-ae301900-5ea3-11eb-8091-161d1e5331c9.png)
how it is build:
![image](https://user-images.githubusercontent.com/3217184/105647116-e46d9880-5ea3-11eb-9ea1-fa03abd5808c.png)
as far as I see the only solution would be to fully remove the vsc-controller div or move the div one div up but then it won't show in the right corner

PR #755, opened 2021-01-24
Add honeycode.aws to blacklist (videospeed mangles the site)

russelldavis:
See #754

I'm not sure what's causing the issue, but this adds the site to the blacklist until it gets fixed.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -19,6 +19,7 @@ var tc = {
       vine.co
       imgur.com
       teams.microsoft.com
+      honeycode.aws
     `.replace(regStrip, ""),
     defaultLogLevel: 4,
     logLevel: 3


Xerothusian:
Would it be unwise to simply allow people to do it themselves?

igrigorik:
Seems to be working a-ok now! :)

russelldavis:
It's still broken.

igrigorik:
Odd, which page / video are you testing?

russelldavis:
Just the home page. The videos are all larger than they should be (for reasons explained in #754), so the page has horizontal scrollbars when it shouldn't (at least on a 15" Macbook Pro screen). The issue is even more obvious if you click the image at the top of the page which shows a larger video player in a lightbox (and two layers of horizontal scrollbars).

Issue #84, opened 2016-02-10
Skipping in netflix causes crash


atlas6523:
Using the skip option, whether in the gui or with the shortcut keys causes the Netflix to go to the "Encountered an Unexpected Error" page.  The code is M7375.  I use the natural/already-provided skipping providing by the arrow keys, so this isn't a problem for me.  I know you said you don't have access to a Netflix account, I just wanted to let you know.

igrigorik:
Ugh. Looks like this might be related to Widevine / DRM: https://help.netflix.com/en/node/14759 - guessing, they may be explicitly disallowing seek's or some such.

igrigorik:
Closing, as I'm unable to reproduce. If only Netflix could sponsor us with a tester account.. :)

rdp:
No work around known yet, unless you want to manually click on some netflix UI ref: http://stackoverflow.com/questions/27927950/controlling-netflix-html5-playback-with-tampermonkey-javascript FWIW :|

Issue #260, opened 2017-09-14
Netflix issue code M7375


Rawsted:
Issue code M7375 appears when i press 
Z for rewind
or
C for Advance

tryed to put some other key bindings but that didn´t work either.

igrigorik:
This is a bug on Netflix side, related discussions:

- https://github.com/igrigorik/videospeed/issues/84
- https://stackoverflow.com/questions/27927950/controlling-netflix-html5-playback-with-tampermonkey-javascript

Unfortunately not much we can do address this on our end.. closing. :(

thefantas:
The addon 'Netflix Dual subtitle for learning languages', manages to make it work. It would be a matter of reviewing how they do it (it works very well on netflix).

stephenbrady:
It looks like this Stackoverflow comment provides a clue about how to resolve: https://stackoverflow.com/a/65488170

Happy to submit a PR if you'll review it.

igrigorik:
@stephenbrady that's a great find! Yes, happy to review if you're willing to pull it together.

PR #781, opened 2021-03-04
Add support for rewind and fast forward in netflix.com

stephenbrady:
Resolve issues #84 and #260

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -801,17 +801,18 @@ function runAction(action, value, e) {
 if(location.hostname == "www.netflix.com") {
   const elt = document.createElement("script");
   elt.innerHTML = "window.addEventListener('message', function(event) {" +
+    "if (event.origin != 'https://www.netflix.com' || event.data.action != 'videospeed-seek' || !event.data.seekMs) { return; };" +
     "const videoPlayer = window.netflix.appContext.state.playerApp.getAPI().videoPlayer;" +
     "const playerSessionId = videoPlayer.getAllPlayerSessionIds()[0];" +
     "const currentTime = videoPlayer.getCurrentTimeBySessionId(playerSessionId);" +
-  "videoPlayer.getVideoPlayerBySessionId(playerSessionId).seek(currentTime + event.data.seekMs)" +
-  "}, false);";
+    "videoPlayer.getVideoPlayerBySessionId(playerSessionId).seek(currentTime + event.data.seekMs);" +
+    "}, false);";
   document.head.appendChild(elt);
 }
 
 function seek(mediaTag, seekSeconds) {
   if (location.hostname == "www.netflix.com") {
-    window.postMessage({action: "seek", seekMs: seekSeconds * 1000})
+    window.postMessage({action: "videospeed-seek", seekMs: seekSeconds * 1000}, "https://www.netflix.com");
   } else {
     mediaTag.currentTime += seekSeconds;
   }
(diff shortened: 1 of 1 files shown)


igrigorik:
Wow, neat-o. Is this a documented / stable API?

stephenbrady:
I'm not aware of it being documented. I'd be surprised if they explicitly support client-side interaction. It seems stable though. I got a clue about the API from this Stackoverflow response from October 2017 (https://stackoverflow.com/questions/42105028/netflix-video-player-in-chrome-how-to-seek). It hasn't changed in four years, so there's that. Data point of one: I've been using it all year on my fork and haven't had any problems.

One caveat is that I added this second commit to the PR to be more defensive about what messages are handled in the presence of other code in the sandbox sending/receiving messages. It might be needlessly defensive, though, and it presumes that Netflix is being served from netflix.com, which I don't know to be true for all geographies. The host checking could cause more problems for users than it's worth - maybe.

pepijnmm:
o nice!

igrigorik:
> It hasn't changed in four years, so there's that. Data point of one: I've been using it all year on my fork and haven't had any problems.

That's more than I hoped for, tbh. On the defensive side: yeah, I think if we want to land this then we might as well accept the risk and make it work across all geographies. This means doing a more relaxed check in hostname and removing the defensive commit. You OK with that?

stephenbrady:
> This means doing a more relaxed check in hostname and removing the defensive commit. You OK with that?

On this point, in isolation, I'm agnostic. Happy to back it. out.

However, I also have this line in there (https://github.com/igrigorik/videospeed/pull/781/files#diff-8dcb72b9c92c5fa9eb8602ce15b90717f8717699b5865db86078df006be5c4c0R801), which limits injecting the message handling script only when it's being loaded on netflix.com. It's not a huge deal if this is being instantiated on every page that videospeed is being loaded into, but Web page bloat is death by a thousand cuts.

So if we remove the hostname checking for the postMessage, then it would only make sense to remove the conditional injection of the script too because if Netflix is ever served from a host other than netflix.com, the end-to-end will still break.

Or we go the other way, and just leave the PR as is and just react if people report Netflix as not being fixed due to different serving domains. My cursory look suggests netflix.com is the only serving host worldwide right now. Besides, rewind and forward with videospeed doesn't work on netflix.com right now anyway, so we're at least giving those videospeed users a fix and everyone else is status quo.

igrigorik:
> So if we remove the hostname checking for the postMessage, then it would only make sense to remove the conditional injection of the script too because if Netflix is ever served from a host other than netflix.com, the end-to-end will still break.

That's not what I was thinking. Rather, relaxing the constraint on matching the suffix. Effectively, if we see a valid `www.netflix....`. Not without risk of false positives but I agree with you that we don't want to blindly insert this on every page.

russelldavis:
It probably makes sense to match on any (or no) subdomain of Netflix as well. So, as a regex, `^(.*\.)?netflix\.[^.]*$`

igrigorik:
^ +1.

pepijnmm:
As far as I know Netflix.com is the universal address in all countries. Urls like Netflix.ru / Netflix.nl get redirected to www.netflix.com .

stephenbrady:
Sadly, this PR may be moot now. Of all things, it looks like the API has changed after four years! Must have happened in the last month or so? I poked around the object graph, but I'm not seeing how to access the player controller object. Unless someone figures that out, it looks like we should close the PR.

igrigorik:
Doh. Thanks for chasing this.. one day!

Issue #789, opened 2021-03-21
The evaluation of Regex expressions in the blacklist is broken


xAdler:
## Expected Behavior
When adding the regex expression `/.*/g` to the blacklist, the VSC should not be initialized on any website.
## Current Behavior
The regex does not match all websites and is still initialized most of the time.
#### Current Implementation
The _RegExp_ is initialized as `var regexp = new RegExp(match);` at **inject.js:391** where _match_ is the string of the regular expression. When we now print the evaluated expression we get `//.*/g/` which already points us to the issue.
#### What is happening?
The _RegExp_ constructor expects to receive only the body of the expression as the first parameter, the flags should be passed as the second argument, which currently results into the expression to be wrapped inside another regular expression without any flags.
#### Current Workaround
Multiple people already found out, that removing the flags fixes the issue. If we remove the flags, but leave the first forward slash, to still indicate, that this entry of the blacklist should be treated as a regular expression, they indeed work properly again.
#### Why does that work?
If, we now print the expression, we get `//.*/` as the result. This is fortunately working fine, as most URLs contain a forward slash after the protocol.
### Possible solution
Before initialization, we have to split the expression into it's body and the flags, so it's properly constructed.
## Migration
In the future we should only allow valid expressions, when saving the options, but we also have to fix the existing ones, so current blacklists still work fine without any modification. 
#### Fixing existing expressions
Fortunately we know, that currently all expressions are working, at least if they were working before, without any flag, so appending a single forward slash, makes them valid. 
#### Migrating them
When saving the current blacklist to the local storage we just update any regular expression, that doesn't offer a trailing slash with none or only valid flags following, by enclosing it in two forward slashes, just like the old implementation did. This makes sure that old expressions don't change behavior. 
#### Risks
There is the small chance, that an expression used as a workaround could pass that test and change behavior, but that's very unlikely.

VimalMollyn:
+1, I have this issue as well. Any idea when it might get fixed, and if I could be of any help?

LeeBuckle:
Hi @xAdler, could you give me an example of your workaround please? Other than knowing what regex does, my knowledge is very lacking so I can only go off stack overflow and testing in a regex program I use.

[...]

PR #790, opened 2021-03-22
The evaluation of Regex expressions in the blacklist is broken

xAdler:
Please view [this issue](https://github.com/igrigorik/videospeed/issues/789) for details.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -1,4 +1,5 @@
 var regStrip = /^[\r\t\f\v ]+|[\r\t\f\v ]+$/gm;
+var regEndsWithFlags = /\/(?!.*(.).*\1)[gimsuy]*$/;
 
 var tc = {
   settings: {
@@ -389,10 +390,15 @@ function isBlacklisted() {
     if (match.startsWith("/")) {
       try {
         var parts = match.split("/");
-        
-        var flags = parts.pop();
-        var regex = parts.slice(1).join("/");
-  
+
+        if (regEndsWithFlags.test(match)) {
+          var flags = parts.pop();
+          var regex = parts.slice(1).join("/");
+        } else {
+          var flags = "";
+          var regex = match;
+        }
+
         var regexp = new RegExp(regex, flags);
       } catch (err) {
         return;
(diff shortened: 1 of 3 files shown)


igrigorik:
@xAdler apologies about the (super) slow feedback.. 

> When adding the regex expression /.*/g to the blacklist

Naive question, but what do the flags offer when we're operating on a single line string? Is the meta problem here that folks are putting in flags and things break because we assume they shouldn't?

xAdler:
> Naive question, but what do the flags offer when we're operating on a single line string? Is the meta problem here that folks are putting in flags and things break because we assume they shouldn't?

The only flag that offers an advantage would be the case-insensitivity, i guess. So I agree, the flags are not necessary, but as they also couldn't break anything, at least to my knowledge, it would be a nice to have, as my implementation proposes.

The meta problem here is, that the **RegExp** constructor expects either a native regex expression as a parameter `new RegExp(/ab+c/i)` or only the body of it `new RegExp('ab+c', 'i')`. As you can see, the trailing and especially the preceding slash is removed, which exposes the meta problem: The preceding slash is required, to tell VSC to handle the line as a regex.

One note regarding my implementation: Currently when saving and the check is finding an invalid regex, it's automatically attempting to fix it. This could be confusing for some users and we should instead notify them about the invalid regex and offer a "fix automatically" button, in my opinion. What do you think?

igrigorik:
Fair enough. I'm generally wary of any unnecessary code without a clear purpose, but I agree that there is a footgun here for some users. Your suggestion on validation and warning makes sense but I'd skip the "fix automatically"... if you made a mistake you're better off figuring out why and avoiding it in the future. 

Are you OK with that / willing to update the implementation?

xAdler:
I have updated the implementation like requested!

Also I think this code change has a very clear purpose, like seen in the comments of [the corresponding issue](https://github.com/igrigorik/videospeed/issues/789). Currently the huge advantage of offering regex expression is crippled, due to the malfunctioning interpretation. As the above given example shows, it's very hard to solve the quite common issue of transforming the blacklist into a logical whitelist, without knowledge about the underlying problem.

I hope you can agree and thank you for taking a look at it!

Issue #839, opened 2021-08-25
I want you to return to the previous UI


Tooka4540:
On youtube, the button is displayed above the video and it is hard to see.

I can speak only a little English. (google translate)

Tooka4540:
![中央](https://user-images.githubusercontent.com/89531682/130812290-8340d846-7670-4097-ad2a-009397fe4102.png)

Yash089610:
Yes, please. It's not the issue of the new UI, but the new way of finding video frame location. Please use the old method as it didn't have this issue.

toppits:
yes, please. It also looks ugly with the huge padding and 100% border radius. Is that everyone does it, not enough?

Longoon12000:
I do not mind the round corners, however for some reason the placement now is not in the top left anymore but somewhere randomly on the video and the icons in the buttons are not centered.
![vsc](https://user-images.githubusercontent.com/51454971/130826748-a6b8590a-b66a-43ea-9c50-2521c61c9a62.png)
Pls fix <3

paulirish:
Yeah I think this is a dupe of #786, that @russelldavis fixed in #854 

so this can be closed.  

(However that fix is only here in the local dev version of the extension. not yet published to CWS)

Issue #786, opened 2021-03-17
controller position botched on youtube


paulirish:
in https://github.com/igrigorik/videospeed/pull/752 i changed how the controller is positioned. 
this change is landed in git, but hasn't gone to CWS yet.

today @pepijnmm  reported a bug about my change:

> @paulirish For me this merge breaks the youtube controller position(just noticed hadn't updated in a while)
> I tested compared to enforce normal white-space rendering for controller (#753) so only this commit changed things
> Before merge:
> <img alt="Screenshot_8" width="235" src="https://user-images.githubusercontent.com/3217184/111382338-272f3e00-86a7-11eb-876a-04b0971c1114.png">
> after merge
> <img alt="Screenshot_7" width="585" src="https://user-images.githubusercontent.com/3217184/111382398-3b733b00-86a7-11eb-9164-f3b03c57d303.png">

i can repro this as well.

fixing this is going to require some more thinking (and some testing..)  

We'll have to find a positioning calculation that's respectful of offsetParent but still gives results across all target sites.

0x7633:
Many players have this problem. Here are 2 more examples:

instagram.com:
![image](https://user-images.githubusercontent.com/52445324/111441528-5f378f00-8742-11eb-9a73-d448f7ab1cab.png)

fox.com.tr:
![image](https://user-images.githubusercontent.com/52445324/111441684-83936b80-8742-11eb-8dff-9b46c5c9289e.png)

pepijnmm:
@igrigorik is it maybe better to reverse https://github.com/igrigorik/videospeed/commit/a25373f1d831fe06430c2e9e87dc1bd1aabd25b1
if it breaks more then it fixes?
if no one will fix it as temp we could do also do if else where if is for hulu right?(for now I will just stay on the previous release.

russelldavis:
Update: there's a simple fix by subtracting the `getBoundingClientRect` of the `offsetParent`. PR here: https://github.com/igrigorik/videospeed/pull/854

PR #845, opened 2021-08-26
fix for hulu fix commit #752

pepijnmm:
fix for commit https://github.com/igrigorik/videospeed/pull/752 which broke youtube and other websites vsc positioning.
Made because the bug  #786 is now in production
 @igrigorik can we make a build with this?
 a lot of people are now complaining about the centering #839 because it is uploaded to the store.


pepijnmm:
@igrigorik if you are busy is there another person that can merge/upload to the store?
every day new issues about this problem get added and most people don't know how to sideload extensions.

Issue #851, opened 2021-08-30
VideoSpeed Controller not appearing in the corner of the screen


rohankirtiraj:
VideoSpeed Controller not appearing in the corner of the screen, after the latest update.
Despite moving it to a corner, it again goes to its original position on the next video.

its position is very irritating.
I do not want to hide it.
I need to see it, in the corner.

russelldavis:
Duplicate of https://github.com/igrigorik/videospeed/issues/786

Issue #850, opened 2021-08-28
(new) Video controller appears near the center of the video, instead of top left


LynxtX:
before the design of the controller changed, it always appeared at the top left,

now, it always gets in the way of the video, i cant always drag it to the top left, its too much trouble
![Uploading Screenshot 2021-08-28 212504.png…]()

lakpro:
Hey @igrigorik @oggy  @paulirish  @jbrains  ,

The top=0 and left=0 were getting overridden by element.style 

I fixed this issue in the following PR : https://github.com/igrigorik/videospeed/pull/852

Kindly merge it.

PS: (tagging everyone) It's my first PR, not sure what else to do !!

russelldavis:
Duplicate of https://github.com/igrigorik/videospeed/issues/786

PR #852, opened 2021-08-31
Reset the position of controller to the top left

lakpro:
The positioning of the controller was getting overridden by the element.style.
Added "!important" to keep the controller in the top left by default.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -273,9 +273,8 @@ function defineVideoController() {
     log("initializeControls Begin", 5);
     const document = this.video.ownerDocument;
     const speed = this.video.playbackRate.toFixed(2);
-    const rect = this.video.getBoundingClientRect();
-    const top = Math.min(rect.top, 0) + "px";
-    const left = Math.min(rect.left, 0) + "px";
+    const top = Math.max(this.video.offsetTop, 0) + "px";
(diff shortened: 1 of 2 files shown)


lakpro:
Hey @igrigorik  @oggy  @paulirish  @jbrains ,

I fixed the following issues : #850 and #851 

Kindly merge this.

PS: (tagging everyone) It's my first PR, not sure what else to do !!

pepijnmm:
image
but the top and left are supposed to be overwritten.
the function to override is in tc.videoController.prototype.initializeControls https://github.com/igrigorik/videospeed/blob/8f3d6b60491d44ecbca4c5ffc9efc1af0a50bd39/inject.js#L272

lakpro:
Hey @igrigorik @oggy @paulirish @jbrains,

My previous code had a bug about dragging. 
I have removed the " !important " tag in my previous code as it did not let the user drag the controller.

There was some issue in inject.js where it was not coming to the corner due to max function, which I changed to min. I have tested this code on a few websites, and it's working fine.

As pointed out by @pepijnmm, the overridden values also get fixed, which were earlier wrong.

Now controller comes to the top left corner, and it is draggable.

I kindly request you to look into this issue asap as many people are facing difficulty and it's annoying everyone.

PR #854, opened 2021-09-01
Fix positioning for videos with non-zero offsetTop/offsetLeft

### LINKED ISSUE #786: controller position botched on youtube

russelldavis:
Fixes #786

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -274,8 +274,12 @@ function defineVideoController() {
     const document = this.video.ownerDocument;
     const speed = this.video.playbackRate.toFixed(2);
     const rect = this.video.getBoundingClientRect();
-    const top = Math.max(rect.top, 0) + "px";
-    const left = Math.max(rect.left, 0) + "px";
+    // getBoundingClientRect is relative to the viewport; style coordinates
+    // are relative to offsetParent, so we adjust for that here. offsetParent
+    // can be null if the video has `display: none` or is not yet in the DOM.
+    const offsetRect = this.video.offsetParent?.getBoundingClientRect();
+    const top = Math.max(rect.top - (offsetRect?.top || 0), 0) + "px";
+    const left = Math.max(rect.left - (offsetRect?.left || 0), 0) + "px";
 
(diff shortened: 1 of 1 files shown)


russelldavis:
Yep I've tested a bunch on Youtube. I don't have a Hulu subscription either, but I've tested on other sites that were affected by the same bug as Hulu (due to a `transform` on the video location), e.g. Reddit. I've also tested on other sites that were affected by #786, e.g. https://mlb.com/news.

Also tested on several other sites in general, e.g. Amazon Prime Video and Vimeo.

pepijnmm:
works good for me too!
for hulu I tried https://www.hulu.com/movie/good-boy trailer (didn't want horror but could not find other trailers XD )

I will close #845 

thanks! Hope they merge it soon.

pepijnmm:
@igrigorik do you have time to check this one out to fix to position of the vsc?

igrigorik (approved):
Apologies for the delay on the review and thank you for the fix! 👍🏻

paulirish:
@russelldavis thanks for fixing this. 🙏 

back when i introduced the bug in #752 i said

> `getBoundingClientRect` takes transforms into account, but otherwise I think it has consistent semantics with `offsetLeft`/`offsetTop`. …

well... really dumb of me not to remember or consider a positioned parent's effect.  

Appreciate you cleaning up my mess. :)

PR #866, opened 2021-09-12
Update to work with the latest Apple TV

nathanielherman:
The old insertion code no longer works, and the change that allows injection into apple-tv-plus-player is a different attribute change than 'aria-hidden'

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -361,8 +361,7 @@ function defineVideoController() {
         p.insertBefore(fragment, p.firstChild);
         break;
       case location.hostname == "tv.apple.com":
-        // insert after parent for correct stacking context
-        this.parent.getRootNode().querySelector(".scrim").prepend(fragment);
+        // pass-through, the latest version of AppleTV works with default behavior
       default:
         // Note: when triggered via a MutationRecord, it's possible that the
         // target is not the immediate parent. This appends the controller as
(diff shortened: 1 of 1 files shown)


inject.js
```diff
@@ -361,8 +361,7 @@ function defineVideoController() {
         p.insertBefore(fragment, p.firstChild);
         break;
       case location.hostname == "tv.apple.com":
-        // insert after parent for correct stacking context
-        this.parent.getRootNode().querySelector(".scrim").prepend(fragment);
+        // pass-through, the latest version of AppleTV works with default behavior
```
igrigorik, line 364: Can we remove this branch? No purpose, right?

inject.js
```diff
@@ -667,7 +667,7 @@ function initializeNow(document) {
     );
   });
   observer.observe(document, {
-    attributeFilter: ["aria-hidden"],
+    attributes: true,
```
igrigorik, line 670: This is a heavy performance tax, instead of opting in wholesale, we can provide a list of attributes. Is there an attribute we can target for AppleTV?

igrigorik (commented):
Thanks for pulling this together! A couple of questions..

inject.js
nathanielherman: sounds good

inject.js
nathanielherman: yes there is, I'm a little worried about it changing cuz it's kind of random, but yeah the perf tax might be high otherwise

inject.js
nathanielherman: I ended up modifying it so that the hover works too (before it would show and you could use hotkeys but not modify the UI directly)

igrigorik:
@nathanielherman looking at your last commits, this looks good.. Good to merge from your side?

igrigorik:
@nathanielherman testing it locally, the controller is injected successfully — yay. However, can't move it, stacking order?

LMK once we're good to merge.

nathanielherman:
@igrigorik can't move it? I can hover it and modify it successfully, I'm not familiar with what moving means though

nathanielherman:
I'm good to merge, probably won't have time to modify it further if there's more you want to do

flubbian:
Hi! Has this issue been fixed? Is it finally working with Apple TV+? Thank you!

igrigorik:
@nathanielherman merging, thanks again for your contribution!

flubbian:
How do I enable it on tv.apple.com? It's still not working.

nathanielherman:
@flubbian you have to install the extension directly from the latest github version rather than through the store, that should work

flubbian:
@nathanielherman OK, but how do I do it exactly? Is there any tutorial?

notsuhas:
image

Hi, Apple TV+ playback speed works but is stuck on the speed set before playing. For instance if speed before starting is 3x and I start playing, it works but I can't reduce speed anymore nor increase and inside the Chrome Extensions I shown the above error

PR #867, opened 2021-09-14
Adopt Manifest V3

dotproto:
This PR updates the extension to use Manifest V3, the latest version fo the Chrome extension platform. This migration was very easy because the changes required were limited to updating names and structures in manifest.json.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,8 +1,9 @@
 {
   "name": "Video Speed Controller",
   "short_name": "videospeed",
-  "version": "0.6.5",
-  "manifest_version": 2,
+  "version": "0.7.0",
+  "manifest_version": 3,
+  "minimum_chrome_version": "89",
   "description": "Speed up, slow down, advance and rewind HTML5 audio/video with shortcuts",
   "homepage_url": "https://github.com/igrigorik/videospeed",
   "icons": {
@@ -15,7 +16,7 @@
     "page": "options.html",
     "open_in_tab": true
   },
-  "browser_action": {
+  "action": {
     "default_icon": {
       "19": "icons/icon19.png",
       "38": "icons/icon38.png",
@@ -37,5 +38,8 @@
       "js": ["inject.js"]
     }
   ],
-  "web_accessible_resources": ["inject.css", "shadow.css"]
+  "web_accessible_resources": [{
+    "resources": ["inject.css", "shadow.css"],
+    "matches": ["http://*/*", "https://*/*", "file:///*"]
+  }]
 }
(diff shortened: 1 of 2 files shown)


manifest.json
```diff
@@ -1,8 +1,9 @@
 {
   "name": "Video Speed Controller",
   "short_name": "videospeed",
-  "version": "0.6.5",
-  "manifest_version": 2,
+  "version": "0.7.0",
+  "manifest_version": 3,
+  "minimum_chrome_version": "89",
```
pepijnmm, line 6: question why do we want to put the minimum chrome version on 89?
Why not keep it so that chrome users from 2020 could even install it.

dotproto: Chrome 89 is the first stable release to support Manifest V3.

igrigorik (commented):
Thanks for pulling this together! Overall, 👍🏻 

Before we merge, are there any behavior/capability changes to watch out for or test, as a consequence of moving to v3?

dotproto:
I don't think there are any other changes that affect this extension, but generally speaking, yes, there a few different changes that extension developers should be aware of with the introduction of the new platform. They're covered in more detail in the [Manifest V3 Overview](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview/), but to borrow from the Feature Summary section of that page: 

> There are a number of new features and functional changes for extensions using MV3:
> 
> - [Service workers](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview#service-workers) replace background pages.
> - [Network request modification](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview#network-request-modification) is now handled with the new [declarativeNetRequest API](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/).
> - [Remotely hosted code](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview#remotely-hosted-code) is no longer allowed; an extension can only execute JavaScript that is included within its package.
> - [Promise](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview#promises) support has been added to many methods, though callbacks are still supported as an alternative. (We will eventually support promises on all appropriate methods.)
> - A number of other, relatively [minor feature changes](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-overview/#other-features) are also introduced in MV3.

You may also want to consult the [Migrating to Manifest V3](https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/) guide for a more directly actionable set of changes.

manifest.json
pepijnmm: A thanks could not find it anywhere.

PR #874, opened 2021-09-29
disable user selection for vsc-controller

vogeljo:
Disables text selection when the user is dragging the controller around (.vsc-controller)


igrigorik (approved):
Embarassingly late, but this is great — thank you!

PR #882, opened 2021-10-14
speed not showing: if (offsetRect===undefined) { document.documentElement.appendChild(fragment)

FuPeiJiang:
speed not showing on https://music.youtube.com/
I hope this fallback works on every website
I like that it's at `top:0px; left:0px;` because it doesn't hide part of the video

can someone explain why
`document.documentElement.appendChild(fragment)` works
but
`this.parent.insertBefore(fragment, this.parent.firstChild);` doesn't ?
I can see that the element IS there. by searching (in chrome inspect element) `"drag"`, but it isn't displayed ?

___
I think `Hide controller by default` should be enabled by default, or else it's obtrusive
and `Controller opacity` at least 0.8 for it to easily be seen in a dark background

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -373,7 +373,11 @@ function defineVideoController() {
         // Note: when triggered via a MutationRecord, it's possible that the
         // target is not the immediate parent. This appends the controller as
         // the first element of the target, which may not be the parent.
-        this.parent.insertBefore(fragment, this.parent.firstChild);
+        if (offsetRect===undefined) {
+          document.documentElement.appendChild(fragment)
+        } else {
+          this.parent.insertBefore(fragment, this.parent.firstChild);
+        }
     }
     return wrapper;
   };


magnus-ISU:
I am maintaining a firefox fork, and am considering merging this. I'm wondering; why did you choose to check for `offsetRect`? Is this something which affects more sites than just music.youtube.com? And is `offsetRect` being undefined something which is particular to music.youtube.com somehow, or is it a general issue which could affect any site?

Otherwise it should be easier and cause less potential side-effects to add a rule into the case statement to apply to music.youtube.com specifically here more easily.

Thanks for the fix. And yeah I agree that `Hide controller by default` is the better default, I have hesitated to change too many defaults so far but I've added several features which should improve that use-case (mousewheel changing of speed mostly) and so I think I might change it.

FuPeiJiang:
code you write today and code you wrote 6 months ago

you can add a comment:
```javascript
if (offsetRect===undefined) { // if it has no where to put it (since offsetRect is a pair of positions)
```


https://github.com/igrigorik/videospeed/blob/caacb45d614db312cf565e5f92e09a14e52ccf62/inject.js#L281
https://github.com/igrigorik/videospeed/blob/caacb45d614db312cf565e5f92e09a14e52ccf62/inject.js#L279-L280

> offsetParent can be null if the video has `display: none` or is not yet in the DOM. 

I don't know how a video you're seeing can have `display:none`. so that should be fixed ?
an audio can though

if it adds it as a child of \<audio\> with `display:none` it will simply not work, even the hotkeys won't work
I'm not sure what I ran into was `display:none`, because I don't think `display:none` should affect the hotkeys

___
hmmm, I should document why this was done (inside the code), and what case it covers, before it gets out of hand, no one can read all the pull requests
oh, link to github issue, that's what they do

magnus-ISU:
Ah, gotcha. I am looking at the code so I can update my fork to version 7.1 from 6.3, and I have solved the same problem `offsetRect` gave a different way, by just setting `top` and `left` to `0px`. This means I don't have an `offsetRect` const at all, so don't need to merge this fix.

Thanks for taking a look at it anyway. If I ever encounter a situation where `0px` seems to give incorrect results, I will come back to this, but I haven't found one yet.

igrigorik:
@magnus-ISU if your solution does not require this branch, would you be willing to port and PR the update here? I'd rather go with a generic solution than patch current behavior.

PR #896, opened 2021-11-26
feat/add-blacklist-option

dwikse:
add blacklist option

### DIFF
--- a/options.js
+++ b/options.js
@@ -22,6 +22,7 @@ var tcDefaults = {
     twitter.com
     imgur.com
     teams.microsoft.com
+    cakap.com
   `.replace(regStrip, "")
 };
 


magnus-ISU:
Why should that website be blacklisted? I see a few videos on the homepage which I would want to watch at a faster speed.

dwikse:
because we have several video courses, 
which when students buy the course ,students should to watch the video until the end (not to be speedup/skipped the videos)

pepijnmm:
I think this can be closed.
If the students watches with a faster speed it's their call. I don't see why we would decide what they can and can't use the extention on.

If people read they also read at different speeds right? I don't see why that would be different for videos. Even audiobooks have speed options in most apps for that same reason. 

The extention is using speed/skip methods that already exist in the browser. Its not like if you block it here they can't just change the speed in element inspect, this request would only makes it so that we get a issue where people will say it doesn't work for website xxx.

ChadBailey:
I agree with the commends of @pepijnmm , however, I also think it would be better to work with websites rather than against. If we don't, all we will have is an arms race between them blocking VSC and us getting around their protection.

Instead, as much as I disagree with the notion of restricting the user's options, I think we will eventually be implementing some APIs one of which will allow websites to request VSC to be disabled. We will see when we get to it...

Irrespective of future plans, I won't be merging this as-is since it wouldn't work as expected anyway. Although the videos are on the website, they are actually hosted on YouTube and we certainly will not be adding YouTube to the blacklist.

Issue #906, opened 2021-12-16
Feature Request: Capital and lowercase input is differentiated


magnus-ISU:
Implemented [here, but I unthinkingly did it for the Firefox port of this repo.](https://github.com/codebicycle/videospeed/pull/168)

The main motivation is that this would make it a lot easier to not conflict with website keybindings I think.

Because I don't know when I'll get around to making the changes necessary to merge with here instead of firefox downstream, I will note the questions I had and the changes my solution made. Hopefully someone can take a look and give comments

This refactors a lot of input code. The way it does so is by mostly ignoring the keyCode attribute of an input event now and instead using the string key which contains more information and in a more human-readable way. Unfortunately, this currently means it breaks any preconfigured options a user has set and users will need to reconfigure it. Assuming this is unacceptable (I assume it is) I can add in a migration strategy; but I need to know how best that done.

I decided the default keybindings are all for lowercase letters; but before they would be either. Is that an acceptable change?

I also removed the old 5.3.0 settings conversion code because its been here long enough I think. In doing so, I was able to simplify some logic. Additionally, vine.co does not really exist anymore so disabling the extension there by default is unneeded I think. Let me know if you want me to revert the vine.co change, that was mostly spur-of-the-moment.

I need testing for numpad keys to see if they still work the same. I wrote this all on a laptop without them. I can do this myself though before I make a PR, just wanted to mention it.

There are several TODOs marked around the code where I wasn't sure what was going on and need review to tell me if my comments are correct or not. They mostly relate to other ways I feel refactoring could be done to reduce redundant or seemingly-unused information.

Additionally, it also resolves #499. This is a change I made on this branch because I changed a lot of code related to settings and didn't want to write it twice.

Issue #881, opened 2021-10-09
Ability to move controller on screen not working anymore


ariiannemay:
Been using the video speed controller for months now and I find it very handy and useful. I especially like the feature where I can move the "controller" bar anywhere on my screen but lately after a certain update it has become fixed on my screen on the upper right corner. I hope we could be able to use the "move the controller anywhere" feature again in future updates.

Issue #860, opened 2021-09-09
Zoom Videos Posted On Facebook


wwongcand:
Some Zoom videos posted on Facebook feeeds do not have the option to use the video speed controller. Is there a fix for this? Thanks!

pepijnmm:
Do you have a video as an example?
the code is open-source so anyone can make a fix for it in theory. But not sure if it will happen it depends.

wwongcand:
Yes. This video is an example - https://www.facebook.com/brian.grasso/videos/343808050826387/. It's in a group so not sure if you'll be able to view it or look into it from just the link.

thewheat:
@wwongcand seems that the link is no longer valid or that the video is set to share to only specific people 😅  Do you have any other example?

![image](https://user-images.githubusercontent.com/892961/133922231-1c818fce-a22f-473e-9b0e-8c776fac967a.png)

PR #907, opened 2021-12-16
Differentiate uppercase and lowercase keyboard shortcuts, make controller size configurable, fix some issues on reddit, and netflix, maybe some more stuff

### LINKED ISSUE #499: Feature Request: Larger buttons on screen
Installed this for seniors, they cannot easily see or click the buttons.

Could there be options for making them bigger? and options for other corners of screen?

Thanks alot.

magnus-ISU:
Fixes #499 #881, my request #906, some issues related to #860 though not for facebook itself I think, as well as [some](https://github.com/codebicycle/videospeed/issues/158) issues [from](https://github.com/codebicycle/videospeed/issues/166) the firefox version.

Sorry for absolutely garbage PR, I did all my proper work on the Firefox version and couldn't figure out how to merge it properly

This refactors a lot of input code. The way it does so is by mostly ignoring the keyCode attribute of an input event now and instead using the string key which contains more information and in a more human-readable way. Unfortunately, this currently means it breaks any preconfigured options a user has set and users will need to reconfigure it. Assuming this is unacceptable (I assume it is) I can add in a migration strategy; but I need to know how best that done.

I decided the default keybindings are all for lowercase letters; but before they would be either. Is that an acceptable change?

I also removed the old 5.3.0 settings conversion code because its been here long enough I think. In doing so, I was able to simplify some logic. Additionally, vine.co does not really exist anymore so disabling the extension there by default is unneeded I think. Let me know if you want me to revert the vine.co change, that was mostly spur-of-the-moment.

I need testing for numpad keys to see if they still work the same. I wrote this all on a laptop without them. I can do this myself though before I get back to you on some of my other questions, its important to mention though I think.

Improves the quality of the code in several places, adding more useful comments and improving code reuse. Creates a file `constants.js` which contains common values between `inject.js` and `options.js` - unfortunately, because user scripts aren't modules, the code has to be duplicated at the top of `inject.js`.

Changes the UI slightly to be able to do #499. I tried to get it as close to the original as possible, but I didn't 100% succeed. Let me know if you think it needs to be changed any more.

There are several TODOs marked around the code where I wasn't sure what was going on and need review to tell me if my comments are correct or not. They mostly relate to other ways I feel refactoring could be done to reduce redundant or seemingly-unused information.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -3,7 +3,7 @@
 export const regStrip = /^[\r\t\f\v ]+|[\r\t\f\v ]+$/gm;
 
 //TODO no idea what this does
-const regEndsWithFlags = /\/(?!.*(.).*\1)[gimsuy]*$/;
+export const regEndsWithFlags = /\/(?!.*(.).*\1)[gimsuy]*$/;
 
 // The default values for each setting
 export const tcDefaults = {
--- a/constants.js
+++ b/constants.js
@@ -17,7 +17,7 @@ export const tcDefaults = {
   forceLastSavedSpeed: false,
   enabled: true,
   controllerOpacity: 0.3,
-  controllerSize: "13px",
+  controllerSize: "14px",
   keyBindings: [
     { action: "display", key: "v", value: 0, force: false, predefined: true },
     { action: "slower", key: "s", value: 0.1, force: false, predefined: true },
--- a/constants.js
+++ b/constants.js
@@ -1,0 +1,36 @@
+//TODO verify accuracy of and fix this comment
+// Strips out whitespace before a line, I think
+export const regStrip = /^[\r\t\f\v ]+|[\r\t\f\v ]+$/gm;
+
+//TODO no idea what this does
+const regEndsWithFlags = /\/(?!.*(.).*\1)[gimsuy]*$/;
+
+// The default values for each setting
+export const tcDefaults = {
+  // The default speed videos should play with
+  speed: 1.0,
+  //TODO is this value used? I don't see why it should be needed, can probably be refactored out?
+  displayKey: "v",
+  rememberSpeed: false,
+  audioBoolean: false,
+  startHidden: false,
+  forceLastSavedSpeed: false,
+  enabled: true,
+  controllerOpacity: 0.3,
+  controllerSize: "13px",
+  keyBindings: [
+    { action: "display", key: "v", value: 0, force: false, predefined: true },
+    { action: "slower", key: "s", value: 0.1, force: false, predefined: true },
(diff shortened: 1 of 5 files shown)


magnus-ISU:
I've added several more improvements to my fork (not in this MR); mouse scrolling support, reloading hotkeys without reloading webpages, some other fixes.

I don't have any interest in chromium, and I'm maintaining my own firefox fork now too, so I don't really care much if this gets merged. However, I'll leave it here in case someone else wants these features.

Thanks for creating the extension!

ChadBailey:
I appreciate your contributions none the less. I am currently working on a refactoring effort in #963 which I hope to be able to expand the compatibility of the extension to include other browsers.

I will leave this here for now without merging, as I'm unable to dedicate the needed time to look through it. In an ideal world, we get everything working cleanly in parent project and you can spend your time helping contribute to it rather than having to maintain your fork.

FedeCuci:
I have tested this fork on Firefox (#973) and it works well! Not sure what else I can do to test it further @ChadBailey but I'd be happy to help where I can. Let me know 👍

ChadBailey:
Thank you for testing! I think I'll try and push to get this merged in a couple of days. Was hoping to hear from @igrigorik with his opinions before opening a PR.

I'll have to do the final cleanup steps I have in the bugs sections before I do of course

igrigorik:
> I decided the default keybindings are all for lowercase letters; but before they would be either. Is that an acceptable change?

Hmm. In practice, I've found that having both caps and lowercase mapped to the same function is a useful feature. For example, on vimeo cap-{CMD} does not trigger their UI logic and acts as a nice workaround. I think my preference would be to keep that behavior as is.

FedeCuci:
> > I decided the default keybindings are all for lowercase letters; but before they would be either. Is that an acceptable change?
> 
> Hmm. In practice, I've found that having both caps and lowercase mapped to the same function is a useful feature. For example, on vimeo cap-{CMD} does not trigger their UI logic and acts as a nice workaround. I think my preference would be to keep that behavior as is.

I agree. I would keep both caps and lowercase mapped to the same function as well as there might be cases in which the user forgets the caps-key on, and might be confused as to why the controls are not working. It would offer more customization but would be harder to configure and probably just complicate things.

A solution to satisfy both approaches is to allow the user to chose what they want through a checkbox for example. However the default should be having both caps and lowercase mapped to the same function imo.

igrigorik:
Closing, https://github.com/igrigorik/videospeed/issues/963 refactor lands bits and pieces of the work here — ty!

PR #932, opened 2022-03-28
Support for video/audio in iframe

blono:
To accommodate video/audio in iframes, I have added the ability to change the playback speed via an interactive context menu.

### DIFF
--- a/background.js
+++ b/background.js
@@ -1,0 +1,60 @@
+chrome.runtime.onInstalled.addListener(() => {
+    chrome.contextMenus.create({
+        id: 'slow-down-01',
+        title: 'Slow down(x0.1)',
+        contexts: ['all']
+    });
+    chrome.contextMenus.create({
+        id: 'slow-down-02',
+        title: 'Slow down(x0.2)',
+        contexts: ['all']
+    });
+    chrome.contextMenus.create({
+        id: 'slow-down-04',
+        title: 'Slow down(x0.4)',
+        contexts: ['all']
+    });
+    chrome.contextMenus.create({
+        id: 'slow-down-06',
+        title: 'Slow down(x0.6)',
+        contexts: ['all']
+    });
+    chrome.contextMenus.create({
+        id: 'slow-down-08',
+        title: 'Slow down(x0.8)',
+        contexts: ['all']
+    });
+
+    chrome.contextMenus.create({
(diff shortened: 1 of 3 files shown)


pepijnmm:
I think the idea is good but why do you use an onInstalled listener?
also can't the code be smaller for example with a array foreach same for the onmessage calls.
is the multiplyPlaybackRate btw needed? as you already check if the audio/video exists so just calling on playbackrate straight away would be easier right?
`const speedMod = (multiplier) => {
	const elements = document.querySelectorAll('video, audio');
	if (elements) {
		for (let i = 0; i < elements.length; ++i) {
			try {
				const element = elements[i];
				if (element != null && media.playbackRate != null) {
					media.playbackRate *= multiplier;
				}
			} catch (e) {
				console.error(e);
			}
		}
	}
}`
Not sure if the if elements is needed as i would expect there to always get a array back filled or empty.

blono:
I'm sorry. You are right, I was aware of the things you pointed out, but I thought they were minor details, so I ended up implementing them now.
The indentation in background.js is also set to 4 characters, and the log output uses console.err instead of log, which is also provided, so I would appreciate it if you could either reject the PR or merge only the necessary parts.

pepijnmm:
> so I would appreciate it if you could either reject the PR or merge only the necessary parts.
I can't do either it's not my repository. I am only giving feedback so that you can implement it cleaner.
Other than that I don't think it's possible to only implement parts. it's or the whole merge request or nothing.

igrigorik:
I like the functionality but agree with @pepijnmm's feedback on the implementation. Do you have bandwidth to update/clean it up?

blono:
Thank you for the fix!

> Do you have bandwidth to update/clean it up?

I'm sorry, I didn't understand this English, so it would be helpful if you could write it in a different way..!
Thank you very much!

blono:
1. Simplified the implementation of the speedMod function in inject.js to reflect the review results.
2. The context menu generation in background.js has not been modified because if it is not executed in the onInstalled event, the menu will be generated many times, resulting in a duplicate menu ID generation error.
Reference: https://developer.chrome.com/docs/extensions/develop/ui/context-menu
https://stackoverflow.com/questions/64318529/cannot-create-item-with-duplicate-context-menu-id-in-extension

Please check it out.

PR #977, opened 2022-10-13
Fix for sling.com and any site that uses document.write

russelldavis:
This extension was broken on sling.com because that site uses document.write, which replaces the document, removing all of our event handlers. The fix is pretty simple since we're already observing mutation events.

You can test it on https://watch.sling.com (no account required).


igrigorik (approved):
👍 thanks for the contribution!

PR #984, opened 2022-11-01
Add support for rewind and fast forward in netflix.com

kevin823lin:
Resolve PR #781 problem
I guess the reason it doesn't work is because of Content Security Policy.
(Or Netflix changed the API and changed it back, but it should be unlikely.)
image

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -828,17 +828,13 @@ function runAction(action, value, e) {
 }
 
 function injectScriptForSite() {
-  let inject = true;
   const elt = document.createElement("script");
-
   switch (true) {
     case location.hostname == "www.netflix.com":
       elt.src= chrome.runtime.getURL('scriptforsite/netflix.js');
       break;
-    default:
-      inject = false;
   }
-  if (inject) {
+  if (elt.src) {
     document.head.appendChild(elt);
   }
 }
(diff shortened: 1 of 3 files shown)


HartS:
I'm curious if this will get merged. @igrigorik  have you been able to look at it?

inject.js
```diff
@@ -825,6 +827,32 @@ function runAction(action, value, e) {
...
+  const elt = document.createElement("script");
+
+  switch (true) {
+    case location.hostname == "www.netflix.com":
+      elt.src= chrome.runtime.getURL('scriptforsite/netflix.js');
+      break;
+    default:
+      inject = false;
+  }
+  if (inject) {
```
igrigorik, line 841: Can we simplify it to perform a check on script.src?

igrigorik:
Apologies for the (very long) delay. Overall, looks good — minor nitpick/question. I've not tested it yet though, can you confirm that it's still working with latest version of Netflix player?

inject.js
kevin823lin: Sure, I've already submitted a new commit.

kevin823lin:
> Apologies for the (very long) delay. Overall, looks good — minor nitpick/question. I've not tested it yet though, can you confirm that it's still working with latest version of Netflix player?

Yes, it still works fine.

igrigorik (approved):
👍 thank you!

jvschiavo:
I'm not sure if it adds anything but I can also confirm that it works. Thank you for your work (both of you)!

PR #993, opened 2022-11-24
Adds short cuts to increase and decrease volume

rahulgolwalkar:
Along with speed control, i preferred having a volume control in my left hand .. 
here is a PR for the same 

Feel free to reject it if you think this wont be used my many


igrigorik:
Thanks for the PR! Most keyboards have "native" volume controls nowadays, I'll close this for now until/unless we get more requests for volume controls.

PR #1032, opened 2023-04-03
Option/Settings refactoring, logging improvements, raw setting editing

mitchcapper:
This has many changes, if any of them are wanted let me know and i can refactor into individual commits / PR's.   Here is a list of the changes made:
https://github.com/mitchcapper/videospeed/blob/various_improvements/ourchanges.md

I wouldn't normally just do a blob PR like this but as many of the changes are interweaved it will take longer form me to separate them out and hope to avoid merge conflicts when combining for a feature-set that may not be desired.


mitchcapper:
Rebased onto master also added firefox support.  If merged into master make sure to update the manifest with your own id.

mitchcapper:
Fixed YT issues where your speed changes would get multiplied due to a shadow video element unless you had "forceLastSavedSpeed".  That setting is now migrated to a new video speed event action enum to add the new "IgnoreWhenVideoZeroSized" which fixes the YT issue without ignoring all.

mitchcapper:
?

PR #1062, opened 2023-09-30
added to README that screen refresh may be neccesary to reflect settings

peno022:
When I first began using videospeed, I was wondered why the new settings didn't take effect even after I modified them on the configuration screen and clicked the Save button. The solution was simple: I just had to refresh the tab where the video was playing. 
It's a minor detail, but I wanted to mention it just in case.


igrigorik:
Thank you, great addition and welcome to the OSS contributor family!

PR #1106, opened 2024-05-05
fix(popup.js): fix `setIcon` error using `action`

bk7312:
Manifest V3 uses `action` instead of `browserAction`. 

Reference: https://developer.chrome.com/docs/extensions/reference/api/action#method-setIcon


igrigorik:
thank you!

Issue #742, opened 2020-12-24
Control of sound volume


StevenPaulJobss:
This App is amazing.
But I hope there is also sound controller because Youtube volumer is sit

pepijnmm:
This extension only changes the speed. I think that it is not in the scope of the project. Rather a separate extension(maybe there already exist some) .
For me YouTube volume changer works great.

StevenPaulJobss:
I use Youtube volume change with key direction up and down. But most time it moves movie frame that’s my problem.

zylstra:
Please accept my only vote as one for allowing volume control via the keyboard.

PR #1107, opened 2024-05-05
feat: add volume control (#742)

bk7312:
Add volume control, able to add shortcut to increase or decrease the volume (optional, not in default).

See issue: https://github.com/igrigorik/videospeed/issues/742

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -830,6 +830,10 @@ function runAction(action, value, e) {
         pause(v);
       } else if (action === "muted") {
         muted(v);
+      } else if (action === "louder") {
+        volumeUp(v, value);
+      } else if (action === "softer") {
+        volumeDown(v, value);
       } else if (action === "mark") {
         setMark(v);
       } else if (action === "jump") {
@@ -873,6 +877,14 @@ function resetSpeed(v, target) {
 
 function muted(v) {
   v.muted = v.muted !== true;
+}
+
+function volumeUp(v, value) {
+  v.volume = Math.min(1, (v.volume + value).toFixed(2));
+}
+
+function volumeDown(v, value) {
+  v.volume = Math.max(0, (v.volume - value).toFixed(2));
 }
 
 function setMark(v) {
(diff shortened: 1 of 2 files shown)


mikepiman:
What number do I enter for "increase volume" and "decrease volume"? 

Before I type something, they both say "value (0.10)", which I do not understand.

Then when I type, the field accepts only: 0, 1, 2, 3,... (i.e., an integer).
Typing 0 causes these key shortcuts to have no effect on volume. 
Typing 1 causes the volume to go essentially to 100% or 0%.

Thanks for any help with this.

bk7312:
You should be entering values between 0 and 1, like 0.05, 0.1 or 0.2. It works like a percentage where 0 means 0% and 1 means 100%. The default 0.1 should be good enough where each increase/decrease will change the volume by 10%.

mikepiman:
Ah, it works great;
the text field does not accept ".05"; it needs the leading 0: "0.05".

mikepiman:
Also,
- [bug]: Often, the volume will go back to its original volume, instantly, about 1-3 minutes after I change it.
- [feature request / issue]: The volume control in the YouTube player doesn't visually change; is it possible for it to change, when we use this extension's volume hotkeys?

bk7312:
Generally, most video streaming websites (like youtube) has their own custom video player to control playback (turn on/off captions, change video quality, prevent you from downloading, etc). This extension works by controlling the video via the [html video tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video) and bypasses whatever custom player the website uses. This works most of the time but there's nothing stopping the website's custom player from overwriting whatever changes that were made by this extension.

If you're using youtube specifically, I would recommend using the up/down arrows to adjust the volume. It's not worth the effort hacking the youtube video player (or any other website's custom player) just to change the volume indication.

mikepiman:
True! But the one benefit that I enjoy anout using this extension's volume keys rather than YouTube's, is that they work even when the keyboard focus is not on the video -- for example when typing comments or clicking around the descriptions or suggested videos.

PR #1114, opened 2024-06-02
make customEvent 'ratechange' work in shadow DOMs

kraig-droid:
VSC uses a customEvent when forceLastSavedSpeed is on. When working on my recent update to handle video elements in shadow DOMs, I didn't notice this customEvent (sorry!) CustomEvents dispatched within a shadow DOM have two issues:

1. customEvents with default options do not 'bubble out' of shadow DOMS -- they bubble to the shadow DOM root and stop
2. once options are changed to allow customEvents to bubble out of shadow DOMs, event.target will refer to the root of the shadow DOM instead of the video element

This PR fixes both of these issues.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -473,7 +473,13 @@ function setupListener() {
         log("Speed event propagation blocked", 4);
         event.stopImmediatePropagation();
       }
-      var video = event.target;
+      /**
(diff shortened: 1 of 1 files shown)


igrigorik (approved):
👍 thank you!

PR #1118, opened 2024-06-19
Add functionality to allow advancing/rewinding the video based on subtitle timestamps (aka you wanna rush through something but want to get all the dialogue feature)

Trannosaur:
I haven't implemented this fully (like options etc.), I just did this for my own private use for a specific site (Dropout).

I haven't got the time to take this to the end, but wanted to document it somewhere - and I couldn't figure out how to just do a diff... on an issue... I dunno, I imagine the best thing to do would be to close the pull request - I'll close it as soon as I create it.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -105,6 +105,20 @@ chrome.storage.sync.get(tc.settings, function (storage) {
       force: false,
       predefined: true
     }); // default: G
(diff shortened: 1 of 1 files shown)


Trannosaur:
also known as adhd super happy fun time feature

PR #1123, opened 2024-07-04
Update popup.html

Abinbn:
Adjusting the HTML for the updated CSS
Added few classes restyled the popup HTML code

### DIFF
--- a/popup.html
+++ b/popup.html
@@ -6,13 +6,15 @@
     <script src="popup.js"></script>
   </head>
   <body>
-    <button id="enable" class="hide">Enable</button>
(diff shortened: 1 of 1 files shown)


mitchcapper:
Nice improvement but comes in quite large, at least on high dpi.  Granted most of it should auto scale but here is default: 
image

Here with some revisions:
image

This seems to be pretty close to the scale the native browser uses for the extensions menu.   The changes I made are in https://github.com/igrigorik/videospeed/pull/1032/commits/64d0fdc20ea9c73eede5aec4c08224a6606a53f3

PR #1142, opened 2024-09-22
adjusting default position for x.com mobile

emem123-co:
the video speed controller on x.com would appear underneath the back button in fullscreen when using a mobile browser that allows extensions (e.g. kiwi browser). 

This patch uses a domain-specific override to shift the default position down so that it is usable in this specific case. It is no longer stuck underneath the back button.

### DIFF
--- a/inject.css
+++ b/inject.css
@@ -23,6 +23,13 @@
   top: 10px;
 }
 
+/* x.com */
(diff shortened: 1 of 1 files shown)


igrigorik:
Sadly, looks like Kiwi browser is no more? :( 

Closing, feel free to reopen if it's still useful in other contexts.

PR #1151, opened 2024-11-06
Add support for rewind and fast forward in drive.google.com

zeolee01:
Google Drive Video Player Rewind and Fast Forward Extension

This functionality enables rewind and fast-forward features for videos played through Google Drive on the web. By injecting custom scripts, we can simulate the behavior of video controls for more flexible and user-friendly navigation. Users can control video playback by a specified number of seconds (or minutes) by rewinding or fast-forwarding, providing better control over video viewing.

Features:

	•	Rewind: Jump backward in time by a specified number of seconds, providing the ability to skip unwanted content.
	•	Fast Forward: Jump forward by a specified number of seconds, enabling users to quickly reach the desired point in the video.
	•	Cross-platform support: Works seamlessly on Google Drive’s video player when accessed through the browser.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -155,9 +155,8 @@ function getKeyBindings(action, what = "value") {
 }
 
 function setKeyBindings(action, value) {
-  tc.settings.keyBindings.find((item) => item.action === action)[
-    "value"
-  ] = value;
+  tc.settings.keyBindings.find((item) => item.action === action)["value"] =
+    value;
 }
 
 function defineVideoController() {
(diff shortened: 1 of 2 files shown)


inject.js
```diff
@@ -871,10 +900,20 @@ function injectScriptForSite() {
...
+      break;
+    case location.hostname === "drive.google.com":
+      window.postMessage(
+        { action: "videospeed-seek", seekMs: seekSeconds * 1000 },
+        "https://drive.google.com"
+      );
       break;
     default:
+      // Directly update the media tag if it's neither Netflix nor Google Drive
       mediaTag.currentTime += seekSeconds;
```
igrigorik, line 917: Hmm, isn't the default branch doing exactly the same thing?

zeolee01: the extension is not working for the videos stored in google drive... so i thought this was the reason...

igrigorik: Hmm, it's literally doing the same operation:

```
videoElement.currentTime += event.data.seekMs / 1000;
```

Something else is amiss here.

PR #1168, opened 2025-01-13
added ability to bind Shift+KEY

NvRKqbbx7Lq2UJfd7gkvjbaFpls7F-AQ3L-alr3:
Modified inject.js and options.js to allow binding key combination with Shift such that "A" and "Shift+A"  can be bound to separate actions. This also works for the Firefox fork.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -584,6 +584,7 @@ function initializeNow(document) {
       "keydown",
       function (event) {
         var keyCode = event.keyCode;
+        var shift = event.shiftKey;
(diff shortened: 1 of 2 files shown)


igrigorik:
The fact that shift-key maps to same function is a feature.. on some players non-Shift is already bound to native action, so this allows one key shortcut where that is not the case, with shift-Key as fallback. Given that this is a long-standing and well-used feature, let's keep it as is.

ty for the pull request!

PR #1171, opened 2025-01-16
Fix #1170: Set Draggable Property of HTML5 Video Containers to False

### LINKED ISSUE #1170: Draggable HTML5 Video Containers Interfere with Controller Drag-and-Drop
The HTML5 Video Containers for YouTube videos are by default set to `draggable="true"`. This interferes with moving the speed box via drag-and-drop – the speed box sometimes moves with cursor after (sometimes not); difficult to rearrange and control. 

https://github.com/user-attachments/assets/4d343781-d054-42dc-9630-ef11f32a1e2c

visyat:
Resets the draggable property of the HTML5 Video containers on YouTube to false (on navigating to a YouTube video). Enables users to move the speed controller easily.

### DIFF
--- a/scriptforsite/youtube.js
+++ b/scriptforsite/youtube.js
@@ -1,0 +1,8 @@
+document.addEventListener('yt-navigate-finish', function(event) {
+    if (!location.pathname.includes('/watch')) { return; };
+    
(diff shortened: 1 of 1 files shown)


igrigorik:
I believe this should already be fixed in master. ty for the pull!

PR #1185, opened 2025-05-02
Fix YouTube subtitles lagging behind

### LINKED ISSUE #537: Youtube subtitle does not speed up
Description:
On youtube, the video speed controller adjustment, and the youtube built-in speed adjustment function are not synced.

That is, when adjusting speed via youtube built-in funcion, video speed controller stays at 1.0x, and the opposite is the same.

Issue:
Youtube subtitle only speeds up when adjusting via built-in speed control, not via video speed controller - i.e., when adjust speed via video speed controller, the subtitle will de-sync with the video.

Possible workaround:
Put a dirty fix to make both option sync, or make video speed controller control the subtitle speed of youtube. (someties you need more than 2x, in lots of the country people don't listen, they watch subtitle even if the content it's their mother tongue)

danielzgtg:
With this, subtitles on YouTube should keep up with up to 8x speed. They should no longer lag behind nor alternate between suddenly slowing down and speeding up.

This works by making YouTube's timers fire earlier / more frequently. This makes YouTube pick up the earlier-than-usual video positions caused by Video Speed Controller, and both refresh earlier and schedule the next timer to refresh earlier. A constant 4x boost is used for simplicity (the new script runs elsewhere in the unisolated MAIN world, so getting settings is hard), and 4x should be enough for most people.

YouTube's own speed should not be set to 0.25, or we get the same behavior as before this PR. The constant 4x should be low enough not to cause lag even with timers firing at 8x the normal 1x speed when YouTube's speed is set to 2x.

A funny side effect of this is that ads are cut a few seconds short.

Fixes #537

### DIFF
--- a/inject_youtube.js
+++ b/inject_youtube.js
@@ -1,0 +1,9 @@
+(() => {
+    const oldSetTimeout = window.setTimeout;
+    window.setTimeout = (f, delay = 0, ...args) => {
+        // Make YouTube update captions 4x as frequent so it can
+        // keep up with Video Speed Controller
+        delay /= 4;
+        return oldSetTimeout(f, delay, ...args);
+    };
+})();
--- a/manifest.json
+++ b/manifest.json
@@ -36,6 +36,11 @@
       ],
       "css": ["inject.css"],
       "js": ["inject.js"]
+    },
+    {
+      "matches": ["https://www.youtube.com/*"],
+      "js": ["inject_youtube.js"],
+      "world": "MAIN"
     }
   ],
   "web_accessible_resources": [{


29988122:
No idea that this will ever be implemented by someone 6 years later.

Thanks for the opensource contribution danielzgtg, and hope that it'll pass the test and be merged soon!

igrigorik:
@danielzgtg could you rebase this against updated code structure?

danielzgtg:
@igrigorik Rebased

I ran into #1227 along the way, but it randomly went away and I retested this to work. Since it happens even without my commit, it should be unrelated.

igrigorik:
I'm having a hard time testing this / verifying it has impact. Any tips?

danielzgtg:
Here are some screen recordings of 
[Cosmos Laundromat, Blender Foundation, CC BY-SA 3.0](https://www.youtube.com/watch?v=Y-rmzh0PI3c). VSC is set to 4x and YouTube is set to 1x.

These should be watched with sound or simultaneously.

# Before

Some subtitles overstay and others are skipped.

[VSC 0.9.1](https://github.com/user-attachments/assets/3299c76a-c0ab-471f-8892-17344980ced0)

# After

All subtitles stay in sync.

[Subt](https://github.com/user-attachments/assets/e37f0886-7131-4f8f-92df-47b0e1d86f8e)

igrigorik:
Interesting, I think I see it now, but only at 4x+ -- does that sound about right? Do you scan videos at that (or higher) speed?

danielzgtg:
On that video, the effect is noticeable even at VSC 3x applied to YouTube 2x: After the long sentence "I'm a sheep minding its own business", the words "And I suggest" are spoken before they appear on screen.

Yes, I watch many videos at 4x, but most videos at 3x. The purpose of me installing this extension is to exceed 2x on YouTube. This is true especially for familiar and slow speakers like Sal Khan. When cramming Khan Academy before exams, I watched new-style videos ([example](https://www.youtube.com/watch?v=Jsiy4TxgIME)) at 6x and old-style videos ([example](https://www.youtube.com/watch?v=jFd-6EPfnec)) at 8x. Subtitles staying in sync lets me not need to rewind if I miss a word.

canarslan12:
@igrigorik hello Ilya, tested it work as expected but currently it change all setTimeout and it might cause another problems.
I figured that only subtitles are not rounded so `(delay%10 != 0)` control will be enought for not breaking anything. 
@danielzgtg can you add that before https://github.com/igrigorik/videospeed/pull/1185/commits/959f833368ffbec9a659d14d22cfb79e6ae7191c#diff-a2950df80365a2998d5503d97c7dd0dd731c78579c3c3716758e1a88a9eb29f6R6

danielzgtg:
@igrigorik Anything that this is still blocked on?

kevin823lin:
> Interesting, I think I see it now, but only at 4x+ -- does that sound about right? Do you scan videos at that (or higher) speed?

I usually adjust playback speed using VSC’s shortcuts rather than YouTube’s own shortcuts or UI controls. As noted in the original issue #537, subtitle desynchronization occurs when the playback speed is changed via VSC instead of YouTube’s native controls. In my experience, this issue is not limited to 4x or higher speeds; it can happen at various speeds whenever VSC is used to control playback.

igrigorik:
Thanks @danielzgtg for the thorough investigation and demo videos, and @canarslan12 and @kevin823lin for testing and follow-up.

We can't merge this as-is — not because the fix doesn't work, but because it's in the wrong layer. A global `setTimeout` monkey-patch with a separate MAIN-world content script bypasses the site handler system, which is exactly where YouTube-specific player behavior should live.

We've just landed a `handleSpeedChange(video, speed)` hook on the site handler API (`master`). The YouTube handler can now override it to notify YouTube's player when VSC changes speed:

```js
handleSpeedChange(video, speed) {
  video.playbackRate = speed;
  const player = document.querySelector('#movie_player');
  if (player?.setPlaybackRate) {
    player.setPlaybackRate(speed);
  }
}
```

This would make YouTube's subtitle scheduler aware of the actual speed — no timer patching, no manifest changes. The open question is whether `setPlaybackRate` clamps to YouTube's native 0.25–2x range. If it does, we'll need to experiment with alternatives (e.g. dispatching YouTube's internal speed-change events, or a more targeted approach in `initialize()`).

@danielzgtg if you'd like to take a crack at this with the new hook, see `docs/custom-player-handler.md` for the guide. Happy to help work through it if you hit the clamping issue.

---
_Generated by [Claude Code](https://claude.ai/code)_

danielzgtg:
I think it ***might*** have worked once on my linked test video, having built at 28f0ef2b0465e582f2aaf3a3d54985829e76171d . Then it stopped working with #1458.

Issue #1152, opened 2024-11-11
Request: Improve expand/collapse of controls overlay for mouse and touch


hounvs:
**Context**
When only using mouse or touch to watch videos, the controls overlay isn't very user-friendly or sometimes isn't functional (due to Windows issue, see below). This means that a keyboard may be effectively required to use the actions.

**Issue**
With a mouse, you need to hover over the smaller target of the current speed and then navigate to the right to click the action you want. If you accidentally move too vertical or horizontal, it collapses and you have to try again. This can be harder when using a touchpad where fine control of the mouse isn't as easy or if you're just not as accurate for whatever other reason.

With a touch device, Windows seems to do weird things with moving your mouse to that location or not. I don't really get it and it's super inconsistent at the OS level. The result is that sometimes tapping the current speed overlay results in it expanding and staying expanded (since it thinks your mouse it there). Other times, it ignores the tap or just flickers the expanded controller (likely because it thinks your mouse is elsewhere). Again, this is a Windows issue so it can't be resolved directly but it stems from the reliance on mouse location as the sole trigger for expanding.

**Resolution**
I'm just going to throw some ideas out there so someone else may have a better approach in mind:
This would be easier if the controls could be resized but it would also be nice to keep it expanded if desired. Perhaps clicking the current speed would trigger expand/collapse so it can be done on the fly while the current auto-expand/collapse remains. And maybe provide an option to configure the delay before collapsing.

**Extra notes**
This is may not be relevant if the control overlay were reworked for #1102 but I think the things I mentioned would be quicker tweaks. Combined with #1115 ([specifically my comment expanding the scope](https://github.com/igrigorik/videospeed/issues/1115#issuecomment-2467038319)), this would fully replace the need for a keyboard.

Issue #1102, opened 2024-04-29
Touch friendly "gestures" (like in NewPipe / VLC mobile)


Dani3I:
There are devices such as "surface pro" tablets which have touchscreen on desktop. There is also fork of this extension for firefox, and it is possible to install firefox extensions on mobile.

In app called NewPipe (for android) when video is in full-screen, touching on right portion of the screen and dragging finger up or down, changes volume (up or down). Touching on left portion of the screen and dragging up or down, changes brightness.
It would be great if this extension would allow to use this or similar gesture to change speed of the video. Simmilarly it also works in mobile version of VLC app.
Another way I thought of is to create some semi transparent button appears on middle left or right side (or other) and touching and dragging it would "reveal" slider to change video speed (and this dragging would change the video speed up or down). It would be kinda simmilar to current semi-transparent button, but it would open slider instead of buttons, and it would appear in different (than upper left) position of the screen. So it could be then made in settings to choose If someone wants buttons or slider; If button should be always visible or auto-hiding; What is the min and max speed. If the button would be only clicked (and not dragged) then slider would also pop up but just clicking in desired position on slider would change video speed (instead of dragging), and then I would argue that it would even benefit desktop users (but In my opinion even dragging would benefit desktop users - I really liked using scrolwheel to change video speed when it was avaiable).

boognish-rising:
Yeah I'm responding from my Android tablet that I have Video Speed Controller installed on via Kiwi Browser. There's also Yandex and all of the many Firefox forks and a handful of other mobile browsers with extension support that could also install Video Speed Controller for use on a touch screen device. So I'm for this but I probably utilize Video Speed Controller on desktop more than anywhere if for no other reason then mobile video players generally have speed controls built in while this is not always the case on desktop so I would just want to make sure that the added functionality didn't detract from the desktop experience. In the case of apps like NewPipe, VLC, YT ReVanced or MX player etc., all of which have the type of touch gestures you're describing, interacting with these apps using a mouse is pretty smooth and intuitive. As long as the same could be said for Video Speed Controller if some this were to be added then I'm all for it

Issue #999, opened 2022-12-11
small buttons


suslik1999:
I use this for Kiwi Browser on android - buttons very small. Can you make some settings for it?

PR #1192, opened 2025-05-16
add option to edit button size for better support of touchscreen.

roiwirven:
I added a new option to configure the "controller size" as the button were way too small to be used with a smartphone or tablet.

I also made it so the buttons catch the touchstart event as without that a double tap on a button would be caught by the video and rewind the video (on youtube).

I did consider adding support for tactile controller repositioning, but while it did work, I didn't manage to override the default touchmove behavior (that on youtube enables to switch to picture on picture mode) so I gave up for that feature that was not very useful anyway.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -321,6 +321,7 @@ function defineVideoController() {
       (e) => {
         runAction(e.target.dataset["action"], false, e);
         e.stopPropagation();
+        e.preventDefault();
       },
       true
     );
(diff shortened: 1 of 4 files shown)


roiwirven:
Fix #1102, fix #1152, fix #999

roiwirven:
I also added a quick change to fix #1170

igrigorik:
👍  and thank you!

PR #1194, opened 2025-05-26
fix prime video black screen

### LINKED ISSUE #1191: The primevideo issue is back
the screen goes black when there is overlay
please fix it

pepijnmm:
fix #1191 #1186 (same problem)
prime made vsc-controller width/height 100% because it's a div and they added CSS for those. 
image
Default CSS parameters have been added to make sure that the same problem doesn't happen again(not counting if the video platform does #id-name > div > div)

If you review this PR: I can change the CSS so that only the width and height will be reset to default(auto). but other video platforms can then still change other lines which will make the video speed controller break again by accident.

To reproduce the error the following video seems free to watch: https://www.primevideo.com/-/us/detail/0TAAO3YTUNN2QPVKG2TTHYAYB6/


igrigorik:
Resolved via https://github.com/igrigorik/videospeed/pull/1188

PR #1199, opened 2025-06-08
Fix #1196: Make add-on overwrite site shortcut by default

### LINKED ISSUE #1196: Time Advance is additive, not overwrite.
If a shortcut overlaps with a site's time advance, the time advance will be the sum of the add-on's settings and the site's settings.

This may be the intended behavior, but it is multiplying the problem of “add-on behavior changes from site to site”.
In this case, there is no point in setting up Time Advance using an add-on.

hevannder:
### Issue #1196

I've read the issue linked and for me it made sense.

When you click to "Show Experimental Features", it shows at the side of each shortcut a select element where you can set it to prevent the default behavior for this shortcut on the website to happen.

```html
<option value="false">Do not disable website key bindings</option>
<option value="true">Disable website key bindings</option>
```

The default value is the 'false' option (first option) and it keeps both behavior happening at the same time.

I don't know if the correct behavior should be this. In my opinion, the default should keep only the add-on behavior, but the option to change could still be available.

Additionally, in my opinion, the text of these options is not very clear, as it does not explicitly state the expected behavior of each option.

### What I did
- Set to true the force attribute of the items present in the tc.settings.keyBindings in order to make it overwrites by default the site's default shortcut for all keys set in the extension.
- Changed the text of the selection options to little bit more clear meaning of the expected behavior.
    ```html
    <option value="true">Override site shortcut</option>
    <option value="false">Combine with site shortcut</option>
    ```
- Created a function to convert the string values from the selection options into booleans. Since the values are read as strings from the DOM, this ensures they are consistently treated as boolean values. This allows the conditionals to use if (item.force) directly, instead of comparing to the string "true"

### Notes
Let me know what do you think about take it out of the group of experimental features and everyting about the changes.

### DIFF
--- a/inject.js
+++ b/inject.js
@@ -68,42 +68,42 @@ chrome.storage.sync.get(tc.settings, function (storage) {
       action: "slower",
       key: Number(storage.slowerKeyCode) || 83,
       value: Number(storage.speedStep) || 0.1,
-      force: false,
+      force: true,
       predefined: true
     }); // default S
     tc.settings.keyBindings.push({
       action: "faster",
       key: Number(storage.fasterKeyCode) || 68,
       value: Number(storage.speedStep) || 0.1,
-      force: false,
+      force: true,
       predefined: true
     }); // default: D
     tc.settings.keyBindings.push({
       action: "rewind",
       key: Number(storage.rewindKeyCode) || 90,
       value: Number(storage.rewindTime) || 10,
-      force: false,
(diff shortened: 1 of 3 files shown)


igrigorik:
Agree the text is not clear, I'll update that. However, I don't think we can make this default behavior, as hijacking site shortcuts will cause a lot of support / compatibility issues.

PR #1235, opened 2025-07-24
Add F13-F24 support for custom key bindings

### LINKED ISSUE #1187: Custom key bindings do not allow for F13-F24 keys to be assigned
Image
In this page, when i try to use any of the F13-F24 keys that i have on my macropad, it simply ignores the keypress and will not assign those keys.

I found this bit of code:
https://github.com/igrigorik/videospeed/blob/d7856864138f95b7acf1a7fdcc96f397c4248502/options.js#L30-L77

Seems like adding F13-F24 should not be a problem

grden:
Made simple changes so that users can F13–F24 keys as custom shortcuts.

- Added F13–F24 key codes to keyCodeAliases in options.js
- Manually tested, confirmed that the key name appears and can be assigned.
 
This closes #1187


igrigorik:
We parallel tracked this one. Fixed in https://github.com/igrigorik/videospeed/commit/41f81432793b226100e7eb97a3a8db1786804cad. Ty for the pull!

PR #1273, opened 2025-08-08
Adds an option to place the controls at different corners of the video frame

dennymarcels:
Originally, the extension places the controls at the top left, but this PR introduces an option in settings to choose from top left, top right, bottom left and bottom right.
If placed on the right side, the controls are mirrored for proper alignment, meaning the controls will expand to the left (instead of the right) with mouse hover.
If placed at the bottom, they will adjust padding/offset to avoid overlaying with the original video controls.
IMPORTANT: I know very little of JS, so all edits were made by Copilot. I did extensive testing in a browser (as suggested in the guidelines) and believe all is fine, but a more detailed revision should be welcome.

### DIFF
--- a/CONTROLLER_POSITION_FEATURE.md
+++ b/CONTROLLER_POSITION_FEATURE.md
@@ -1,0 +1,96 @@
+# Controller Position Feature
+
+## Overview
+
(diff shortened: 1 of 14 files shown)


dennymarcels:
I will PR this feature again with some other fixes

Issue #1272, opened 2025-08-08
New update broke default speed


inxendere:
For some reason now, seemingly randomly, whenever I click on a new video on youtube or twitter or anywhere else, the speed defaults to the lowest, 0.07. This never happened before this new update. pls fix.

Fasciotti:
It's happening to me as well. So far It is affecting only Udemy and Crunchyroll for me.

georgeyu26:
To revert back to version 0.7.4 when everything was working fine, follow the instructions in this link:
https://github.com/igrigorik/videospeed/issues/1257#issuecomment-3140619025

daniel-d7a:
I forked the repo trying to trace it, it happens when seeking the video to some point that hasn't loaded it dispatches an event setting the video speed with 
`video.playbackRate` which happens to be 0 for some reason, then it checks the speed and if it is less than 0.07 it sets it to 0.07

I will try to fix it and see if I can make a pull request

Raunak-Kumar7:
I'm also facing this issue. For now moving to 0.7.4 works for me. Thanks @georgeyu26

ChuckinAllen:
I also have the same issue.

RishabSwift:
Same issue here

BC176:
Also happening to me on Udemy, for visibility

daniel-d7a:
Guys I pushed a pull request that solves the issue, how to get it noticed and merged faster ?

georgeyu26:
> Guys I pushed a pull request that solves the issue, how to get it noticed and merged faster ?

@daniel-d7a you can try reaching out to @igrigorik, the owner of this repository. For now, I'd recommend everyone revert back to version 0.7.4 (instructions linked above).

igrigorik:
Duplicate, please track https://github.com/igrigorik/videospeed/issues/1262

PR #1275, opened 2025-08-09
fix: fixed issue where the video speed goes down to 0.07

daniel-d7a:
fixes issue #1272

hey this is my first time contributing to open source.
I love this project and have been using it for a 3+ years now, it would be great if I can give back to it.

please tell me if my solution is good or if it need s any adjustments
would be very glad to hear any feedback you have <3

### DIFF
--- a/src/utils/event-manager.js
+++ b/src/utils/event-manager.js
@@ -186,6 +186,19 @@ class EventManager {
       return;
     }
 
+    // the speed is too low, propably a bug
+    // however it shouldn't happen
+    /*
+     * this solves a bug where a weird event gets dispatched that
+     * has detail that like this
+     * {origin: "videoSpeed", speed: "0.07", source: "external"}
+     * that wasn't getting caught by the other conditions
+     */
+    if (event.detail && Number(event.detail.speed) < 0.1) {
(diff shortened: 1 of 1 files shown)


daniel-d7a:
to explain what happens it is mostly ok
but seeking the video dispatches an external event from 'videoSpeed' with a speed a 0.07 (aka min speed)
and I know there is a condition to handle events from 'videoSpeed' but for some reason it doesn't catch this one, and I can't figure  where the event is coming from
I tried multiple websites such as sharepoint and facebook and they both had the same problem, and others pointed out the same problem on other websites so it is not the website

I added a small condition that dismisses any change with speed less than 0.1 and that seemed enough to solve it and pass all the tests

daniel-d7a:
Hi, can you have a quick look here @igrigorik ?

inxendere:
bump

igrigorik:
Thanks guys. This should no longer be necessary with fixes in: 
- https://github.com/igrigorik/videospeed/commit/fc71d482e687f643042fbcdfd381a5c9ae516e28
- https://github.com/igrigorik/videospeed/commit/31fc5f5509320ba43b1852181b619373de92bec9

PR #1285, opened 2025-08-16
Adds an option to place the controls at different corners of the video frame

dennymarcels:
*** I am creating a new PR because I pruned the edits from the previous PR, to add as minimum edits as possible. ***
Originally, the extension places the controls at the top left, but this PR introduces an option in settings to choose from top left, top right, bottom left and bottom right.
If placed on the right side, the controls are mirrored for proper alignment, meaning the controls will expand to the left (instead of the right) with mouse hover.
If placed at the bottom, they will adjust padding/offset to avoid overlaying with the original video controls.
Another feature is to show the controls only on mouse hover, this can be enabled in settings.
I added a video showing how the new features work. (The first time using bottom-right, I was too quick to reload the page, then it didn't work at first.)
IMPORTANT: I know very little of JS, so all edits were made by Copilot. I did extensive testing in a browser (as suggested in the guidelines) and believe all is fine, but a more detailed revision should be welcome.

https://github.com/user-attachments/assets/f099b9c7-17e4-48ef-b00f-5dab8f1075e2

### DIFF
--- a/CONTROLLER_POSITION_IMPLEMENTATION.md
+++ b/CONTROLLER_POSITION_IMPLEMENTATION.md
@@ -1,0 +1,53 @@
+# Controller Position Implementation Summary
+
+This implementation adds essential controller position functionality to the `control-pos` branch, copying only the necessary code from the `control-position` branch.
+
+## Features Added:
+
+### 1. Controller Position Settings
+- **controllerPosition**: String setting with options: 'top-left', 'top-right', 'bottom-left', 'bottom-right'
+- **controllerHover**: Boolean setting for hover-only display (secondary trait)
+
+### 2. Position Configuration System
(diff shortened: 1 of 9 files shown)


igrigorik:
Superseded by the user-editable Controller CSS feature (bd44168), which provides the same capability (and more) through CSS rules rather than a settings dropdown.

Users can position the controller at any corner, per site, via Options > Advanced > Controller CSS:

Bottom-right globally:
`vsc-controller { top: auto; bottom: 10px; left: auto; right: 10px; }`

Top-right on a specific site:
`:root[style*='--vsc-domain: "app.plex.tv"'] vsc-controller { left: auto; right: 10px; }`

Closing in favor of the more general solution. Thank you for the PR — the corner-positioning use case was a key motivator for making controller CSS user-editable.

Issue #1291, opened 2025-08-17
When clicked on a part of a udemy course video , speed automatically gets set to 0.7 speed.


Polly333:
I would love to solve this issue and contribute to this tool. Please let me know the starting points and tech revolving setting the speed.

Polly333:
Is node.js used ?

Z3RO-O:
I am also facing this issue, also the current tab which is opened should save the last used speed.

lancerharris:
This also appeared for me about a week or two ago. was working fine before that. It also sort of randomly gets set back to .07 when rewinding 10 seconds.

saman-waruka:
I’m experiencing the same issue.  
The playback speed keeps defaulting to 0.07 whenever a new video starts or when I scrub forward/backward.

For more infomation:
First time video loading I see play back speed is 1.0 and quickly change to 0.07 when video played.

igrigorik:
Should be fixed in 0.9.3 -- out now.

Issue #1293, opened 2025-08-18
Still defaulting to 0.07, and now resetting itself to that on video navigation


BC176:
Originally this was only in Udemy, but now it's happening in other video's as well, an unlisted Vimeo for instance. 

A new video starts and it defaults to 0.07 speed. If I navigate backwards or forwards via the scrubber, it resets to 0.07 speed. 

Screen refreshes will also intermittently reset it to 0.07 speed regardless of being on YT or FB or any site, seems to be the refresh action not the location that causes that one.

saman-waruka:
I’m experiencing the same issue, but so far only on Udemy and Udemy Business.  
The playback speed keeps defaulting to 0.07 whenever a new video starts or when I scrub forward/backward.

For more infomation:
First time video loading I see play back speed is 1.0 and quickly change to 0.07 when video played.

sabrysm:
Same issue happens to me

igrigorik:
Should be fixed in 0.9.3 -- out now.

2ksivaram-cmd:
This still occurs with 0.9.3 verion and with sharepoint videos in microsoft one drive

sabrysm:
> Should be fixed in 0.9.3 -- out now.

The issue stills occurs in Udemy

scriptswhisper:
The issue is still present in v.0.9.5. It always resets to the default value of 0.07, even after changing it in the settings. 
That’s really annoying

PR #1294, opened 2025-08-19
[fix/issue-1291-default-to-x0.07] fix speed defaulting to 0.07

moutasem-isentemiz:
This PR fixes issues #1291 and #1293 where those platforms by default set the video playback rate to 0, resulting in using the `window.VSC.Constants.SPEED_LIMITS.MIN` 
Instead this checks if the `playbackrate = 0` it uses 1 instead

### DIFF
--- a/src/core/action-handler.js
+++ b/src/core/action-handler.js
@@ -383,7 +383,7 @@ class ActionHandler {
 
     // Clamp to valid range
     targetSpeed = Math.min(
-      Math.max(targetSpeed, window.VSC.Constants.SPEED_LIMITS.MIN),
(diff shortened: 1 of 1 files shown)


sabrysm (approved):
LGTM

sabrysm:
> fixed the issue for me

Are you self-hosting the extension?

moutasem-isentemiz:
> > fixed the issue for me
> 
> Are you self-hosting the extension?

i think loading it unpacked, this is how i tried it at least

igrigorik:
Should be fixed in 0.9.3 -- out now.

moutasem-isentemiz:
> Should be fixed in 0.9.3 -- out now.

Just tested the version you mentioned, it doesn't work unfortunately so the issue remains

PR #1310, opened 2025-08-27
Set speed to fast value if reset is pressed when the speed hasn't previously been changed

r3dbaba:
This maintains the original behavior, which is setting the speed to the fast value.

I couldn't find anywhere that the Reset value set in settings is actually used.


igrigorik:
Sorry, missed this earlier. I believe this should be fixed via https://github.com/igrigorik/videospeed/commit/3886fd4967b19cdb42a33021061930fc30d00604. 

ty for the PR!

PR #1312, opened 2025-08-27
Add controller positioning options

Ronco75:
## Description
Adds controller positioning options to allow users to choose where the speed controller appears on videos.

## Changes
- Added `controllerPosition` setting with 6 options: top-left, top-center, top-right, bottom-left, bottom-center, bottom-right  
- Updated options UI with dropdown in advanced settings
- Modified positioning logic to calculate correct positions for all options
- All existing tests pass

## Fixes
Resolves the issue where the controller overlaps with video player controls (like Plex minimize button) by allowing users to position it elsewhere.

## Testing
- Tested all 6 positioning options on multiple video sites
- Verified settings persist across browser sessions
- All existing tests continue to pass

### DIFF
--- a/src/core/settings.js
+++ b/src/core/settings.js
@@ -42,6 +42,7 @@ if (!window.VSC.VideoSpeedConfig) {
         this.settings.startHidden = Boolean(storage.startHidden);
         this.settings.controllerOpacity = Number(storage.controllerOpacity);
         this.settings.controllerButtonSize = Number(storage.controllerButtonSize);
+        this.settings.controllerPosition = String(storage.controllerPosition || 'top-left');
         this.settings.blacklist = String(storage.blacklist);
         this.settings.logLevel = Number(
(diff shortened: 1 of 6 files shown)


igrigorik:
Superseded by the user-editable Controller CSS feature (bd44168).

Users can now customize controller positioning per site via Options > Advanced > Controller CSS. This is more flexible than a dropdown with fixed positions — users can set exact offsets, use different positions per site, and the CSS is live-updated without page reload.

Example — bottom-right on all sites:
`vsc-controller { top: auto; bottom: 10px; left: auto; right: 10px; }`

Example — top-right on Plex only:
`:root[style*='--vsc-domain: "app.plex.tv"'] vsc-controller { left: auto; right: 10px; }`

Closing in favor of the more general solution. Thank you for the contribution — the use case directly informed the design.

PR #1314, opened 2025-08-29
update CONTRIBUTING.md to include build steps

rubywwwilde:
Current CONTRIBUTING guide tells to load the repo root as an unpacked  extension, which results in errors ("Could not load manifest",  "Could not load background script"). This confused me a lot when I tried to install the plugin.

This update adds explicit steps to run `npm install` and  `npm run build` before loading, and clarifies that the `dist/` folder  (the build output) must be selected.


rubywwwilde:
Thank you for the wonderful extension, btw!! It has helped me in the past a lot.

Issue #1341, opened 2025-09-23
Speed is always 1.9 by default


friedpuppet:
Can't change it in settings, nothing obvious helps

excellent-dai:
I also encountered the same situation. The default playback speed of videos is 1.5x, whether I watch short videos or long videos on the website. I would like to ask how to fix this issue.

lavedon:
I am having the same issue - but its 2.20 in my case.

nWestie:
this appears to be the same issue as https://github.com/igrigorik/videospeed/issues/1324#issuecomment-3288484776, see the solution there

bcdax110:
I also encountered this. The default playback speed of videos is 3.0

bcdax110:
> I also encountered this. The default playback speed of videos is 3.0我也遇到过这种情况。视频的默认播放速度为 3.0

I tried the method mentioned in this [comment](https://github.com/igrigorik/videospeed/issues/1324#issuecomment-3288484776), and it worked for me.

igrigorik:
Fixed by the settings storage race condition fix in `93bb729` (on master). Please update and reopen if the issue persists.

Issue #1328, opened 2025-09-12
every video starts at 1.03 for me


gael-cases:
don't know how to set this to 1 when videos begin

666AT9:
The same, it starts at 1,2 speed for me, very annoying!

gael-cases:
I fixed it, You have to go to your settings and put these : 

<img width="468" height="245" alt="Image" src="https://github.com/user-attachments/assets/fbc31508-aa4d-41a4-b88a-af66cfa13e07" />

then when you're on a video, you reset the speed to 1 with R, and refresh the page and it's fixed. <3

abg88:
> I fixed it, You have to go to your settings and put these :
> 
[...]

Issue #1327, opened 2025-09-11
Default playback speed stuck at 4.00


Erendis42:
...since latest update. Very annoying.

0xAryanShah:
yeag speed stuck at a certain speed very annoyiong

abg88:
For me it's 1.5. And there's no relation to preferred speed. Annoying bug.

Erendis42:
then it probably is a speed used when the update happens

Rosaldivo:
F12 -> Application Tab -> Extension Storage -> "Video Speed Controls" -> Sync -> lastSpeed = 1

0xAryanShah:
> F12 -> Application Tab -> Extension Storage -> "Video Speed Controls" -> Sync -> lastSpeed = 1

thanks @Rosaldivo Works like a charm.

Erendis42:
Thanks for helping!

Issue #1324, opened 2025-09-10
Playback Starts at Preferred Speed As Opposed to Default 1x


niruttitime:
After yesterday (seems like the problem is with v0.9.5), playback keep starting at "Preferred Speed" as opposed to normally starting at 1x and  preferred speed is toggle when I hit 'G'.

garyking:
The issue is similar for me, except that every video, at least on YouTube, starts at 1.75x speed rather than the usual 1.0x speed. My preferred speed is 1.5x and my reset speed is 1.0x.

T3Kgamer:
For me it's been starting at 3.10x my preferred is set to 2.75x, this issue persists through a relaunch of chrome

deliro:
Same here. Even video previews (on youtube, for example) play with preferred speed

pedro-h-marchesi:
I've been having a similar issue as well. My preferred speed was set to 1.8x, but every video was opening in 3.5x speed. I tried changing preferred speed, then changing it back to what it was, but it didn't do the trick, videos are always opening in 3.5x speed.

timirons:
Same here, but 2.5x
I think it was the first speed I used after the update.
Also keys don't work until I pause the video.

LewandowskiXYZ:
I have the same issue. It always defaults to the value 2.0, which was my preferred speed. I changed it to 1.0, but the problem persists.

In the commits I see many changes related to speed. Maybe the wrong variable is being used somewhere?

gravitydeepimpactsun:
It's really annoying please fix!

Quas-r:
I have the same issue and mine starts at 2.0x speed which is my preferred speed. Normaly I can undo the speed change by hitting the preferred speed button again but for this case, I can't. I have no idea why the speed starts at 2.0x but for the button not working, I have a theory.

This is probably because the undo system uses a stack and the stack starts empty. Therefore, pressing the preferred speed button sets the speed to 2.0x and adds the change to the stack. If I press it again, lets's say the change gets popped back out and is undone but the previous speed was already 2.0x so there's no effective change.

Let's say the reset button is assigned "R" and the preferred speed button is assigned "G". If you do G + G, no speed change. If you do R + G + G, it works. If you do R + R + G, however, the last G doesn't do anything. That's because the second reset is just an undo for the previous one, which sets the speed to the starting value which is 2.0x in my case. This example supports my theory.

Plz fix it's so annoying, thanks!

Quas-r:
[...]

PR #1348, opened 2025-09-30
lastSpeed setting only used when rememberSpeed is enabled

nWestie:
when loading settings, lastSpeed will be set to 1 if rememberSpeed is not set, regardless of what is stored. This should fix issues where people not using 'remember speed' have their videos always starting at whatever is saved to lastSpeed, but lastSpeed never updates. See https://github.com/igrigorik/videospeed/issues/1341, https://github.com/igrigorik/videospeed/issues/1328, https://github.com/igrigorik/videospeed/issues/1324, https://github.com/igrigorik/videospeed/issues/1327/

Note: I was not able to build the project to test it, I was having trouble getting it working on Windows. If someone can build it to confirm it works  that would be great, or give me pointers on building it - I don't have a huge amount of experience with node. Below is the issue it's throwing:
```
PS C:\Users\westn\Documents\code\videospeed> npm run build

> video-speed-controller@0.8.0 prebuild
> rm -rf dist

'rm' is not recognized as an internal or external command,
operable program or batch file.
```
I'm on windows 11 - like I said I'll keep looking into it to see if I can get a build working. Also tried WSL, that had different issues.

### DIFF
--- a/src/core/settings.js
+++ b/src/core/settings.js
@@ -33,7 +33,6 @@ if (!window.VSC.VideoSpeedConfig) {
         }
 
         // Apply loaded settings
-        this.settings.lastSpeed = Number(storage.lastSpeed);
         this.settings.displayKeyCode = Number(storage.displayKeyCode);
         this.settings.rememberSpeed = Boolean(storage.rememberSpeed);
         this.settings.forceLastSavedSpeed = Boolean(storage.forceLastSavedSpeed);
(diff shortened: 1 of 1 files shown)


igrigorik:
Thanks @nWestie — this is the same rememberSpeed/lastSpeed interaction bug. You correctly identified that `lastSpeed` should only be consulted when `rememberSpeed` is enabled.

This is addressed in #1430 (WIP) alongside the broader settings race condition fix. The refactored `getTargetSpeed()` gates persistence on the `rememberSpeed` flag as you proposed.

Closing as superseded by #1430 — thank you.

PR #1397, opened 2025-12-26
Fix getTargetSpeed initialization when rememberSpeed is off

benzeneringlq:
This PR fixes the issue where videos always start at the lastSpeed (e.g., 1.8x) even when the "Remember playback speed" setting is turned off.

When rememberSpeed is disabled, getTargetSpeed should strictly return 1.0.

### DIFF
--- a/src/core/video-controller.js
+++ b/src/core/video-controller.js
@@ -75,12 +75,14 @@ class VideoController {
   getTargetSpeed(media = this.video) {
     // Always start with current preferred speed (lastSpeed)
     // The difference is whether changes get saved back to lastSpeed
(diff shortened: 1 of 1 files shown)


igrigorik:
Thanks @benzeneringlq — this was a real bug. When `rememberSpeed` is off, `getTargetSpeed()` shouldn't return a stale cached speed.

This is addressed in #1430 (WIP) as part of a broader speed enforcement fix. The architecture now always maintains `lastSpeed` in memory for session use, with storage persistence gated by the `rememberSpeed` setting. New videos get the correct target speed regardless of the persistence setting.

Closing as superseded by #1430 — appreciate the contribution.

PR #1410, opened 2026-02-05
feat: update extension badge to show current playback speed

tomas-barros1:
- Add updateSpeedBadge function to background.js
- Send SPEED_CHANGE message from action-handler when speed changes
- Notify badge on video controller initialization
- Badge shows compact format: '2' for 2x, '1.5' for 1.5x, empty for 1x

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -157,6 +157,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -180,6 +181,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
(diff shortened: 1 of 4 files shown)


igrigorik:
Thanks for the contribution! We've actually explored badge-based speed display before and ended up removing it because it gets really tricky in practice:

- **Multiple videos per tab**: If a page has more than one video (ads, PiP, embeds), each one fires `SPEED_CHANGE` and the badge becomes a race condition — whichever video reports last "wins."
- **Tab switching**: There's no `chrome.tabs.onActivated` listener, so the badge shows stale data from the last tab that sent a message, not the currently focused tab.
- **Navigation cleanup**: Navigating away from a video page leaves the old speed badge showing since there's no teardown/reset logic.
- **No arbitration for "which video matters"**: Deciding which video's speed to display when multiple are present is a UX design problem that doesn't have an obvious answer.

Properly solving all of this requires tracking per-tab, per-video state in the background script, handling tab focus changes, cleanup on navigation, and deciding on an arbitration strategy — which is a significantly larger scope than what's here.

We're going to pass on this for now. Appreciate the effort though!

PR #1412, opened 2026-02-10
Add optional snap-to-1.0x feature

samyk:
If going to YouTube and using YouTube's player to increase speed to 1.25x, then watching something on Netflix at same speed, decrementing by 0.1x repeatedly will never get you to 1.0x. This feature will snap to 1.0x if ever crossing the boundary (1.25x->1.15x->1.05x->[snap]1.0x->0.9x->...)

Resetting is also possible but it's convenient to not have to learn a reset hotkey and use only increment/decrement speeds.

The feature can be disabled from the Options area.

### DIFF
--- a/src/core/action-handler.js
+++ b/src/core/action-handler.js
@@ -393,6 +393,14 @@ class ActionHandler {
       // For relative changes, add to current speed
       const currentSpeed = video.playbackRate < 0.1 ? 0.0 : video.playbackRate;
       targetSpeed = currentSpeed + value;
+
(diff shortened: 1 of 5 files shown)


igrigorik:
Thanks for the idea! We like the snap-to-1.0x behavior but landed a simpler version without the toggle — just always-on, 4 lines in action-handler.js. One extra keypress to get past 1.0 if you don't want it, which feels like the right tradeoff vs. adding a setting.

Closing in favor of our implementation, but appreciate you surfacing this!

Issue #1413, opened 2026-02-10
Bug: settings overwritten from page with stale settings


samyk:
If the Options page is open, there is a race condition of sorts where data gets overwritten by a page with stale settings when triggering speed changes.

Steps to reproduce:
1. open a window with a video so VSC is loaded
2. open the VSC options window in another tab
3. change a setting
4. click Save
5. go back to the first tab that has VSC loaded
6. make a change like increasing the video speed
7. go back to the Options window
8. refresh a few times
9. notice the saved setting is reverted

PR #1414, opened 2026-02-10
Fix setting overwrite race condition

samyk:
Fixes issue #1413. Note this is branched off of the snap-to-1.0x optional feature.

### DIFF
--- a/src/content/inject.js
+++ b/src/content/inject.js
@@ -48,6 +48,9 @@ class VideoSpeedExtension {
         return;
       }
 
+      // Listen for settings changes from options page / popup
+      this.setupStorageListener();
+
       // Initialize site handler
       this.siteHandlerManager.initialize(document);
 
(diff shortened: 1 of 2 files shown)


samyk:
This improvement also has the nice feature of updating VSC settings across all tabs once saved, where previously saving a setting required a refresh on those tabs for the updated setting to go into effect.

igrigorik:
Thanks for identifying this, @samyk — you nailed the root cause. The full-blob overwrite race was real and nasty.

We're tracking a comprehensive fix in #1430 (WIP) using the same approach: `save()` now writes only changed keys instead of the entire settings blob. That branch also addresses the cooldown timing race and fight detection for sites that reset speed.

Your snap-to-1.0x feature lives separately in #1412 and is still being tracked there. One idea from your PR worth revisiting as a follow-up: the `_setupStorageListener()` for hot-reloading settings across tabs — that's complementary to the write fix.

Closing as superseded by #1430 — thank you for the contribution.

Issue #1408, opened 2026-01-26
Youtube max speed before failing lowered from 6 to 4


Visca87:
Hi, I use this extension to skip the boring parts of videos.

I used to have the preferred speed at 6, which beyond that point you lose audio. But recently youtube had an interface update, and on several videos (like 30%?) if I accelerate the video it goes down to like 1fps or less (at enhanced speed), and eventually the video freezes and takes like a minute to recover, or until you reload the page.

I use firefox, I tried disabling and enabling again hardware acceleration, and nothing helps. Videos on reddit still accelerate to x6 satisfactory. I Noticed that yt now allows up to x4 speed to premium users, and maybe that has something to do (I'm not premium, and I can accelerate up to x4 for free). I use like 10 extensions, but I didn't change anything recently.

EDIT: Sorry for not updating before closure, but I got to troubleshoot it. It was a GPU bottleneck. With the change on UI from youtube, it somehow started to recommend higher quality than I was used to. My GPU can only handle 720p@60fps x6 acceleration, but I was getting defaulted to 1440p@60fps on some of the videos the resolution existed, and the limit on those videos for my GPU was x4

igrigorik:
Can't reproduce with lat est build from master, please try 0.9.6+ once it's out.

PR #1415, opened 2026-02-12
feat(popup): add speed presets up to 6.0x

Hexi1997:


igrigorik:
Thanks for the contribution! However, we're going to pass on this one.

Adding 17 more hardcoded preset buttons makes the popup quite crowded (24 total) without a clear UX benefit. Most browsers also lose audio above ~4x, and YouTube specifically breaks above 4x (#1408), so presets up to 6.0x would be misleading for many users.

The existing keyboard shortcuts and +/− buttons already let users reach any speed. If we were to expand presets in the future, the right approach would be user-configurable presets rather than hardcoding more values.

Closing as won't-fix — appreciate the effort though!

PR #1495, opened 2026-04-08
fix: prevent keyboard relative speed baseline reset after tab idle

### LINKED ISSUE #1494: Regression: keyboard s/d shortcuts use 1.0x baseline after tab backgrounding
## Bug Summary
When `rememberSpeed` is enabled, keyboard relative speed shortcuts (`s` / `d`) can apply from a `1.0x` baseline after the tab is backgrounded, instead of using the active/effective preferred speed (e.g. `1.8x`).
This causes unexpected jumps such as `1.1x` or `0.9x` on first key press after returning to the tab.
## Version
Observed in latest release (`0.10.1`).  
Did not observe this behavior previously in `0.10.0`.
## Environment
- Browser: Google Chrome, Brave
- Sites reproduced on: YouTube and other pages with multiple `<video>` elements
- Setting: `rememberSpeed = true`

## Steps to Reproduce
1. Enable `rememberSpeed`.
2. Open YouTube homepage(or another site with multiple `<video>` elements).
3. Play any video.
4. Set preferred speed via keyboard (example: `g` to `1.8x`).
5. Switch to another tab/window and let the video continue in background.
6. Return to the tab.
7. Press `d` (or `s`) once.

## Actual Behavior
Relative keyboard speed change is computed from `1.0x` rather than current/preferred speed.
Example:
- expected baseline `1.8x`
- pressing `d` can produce `1.1x` (or `s` can produce `0.9x`)
## Expected Behavior
Keyboard relative changes should apply from the active video’s effective current speed.
Example:
- from `1.8x`, `d` should produce `1.9x`
- from `1.8x`, `s` should produce `1.7x`
## Important Comparison
UI widget `+ / -` buttons behave correctly.  
The regression appears specific to keyboard relative shortcuts.
## Additional Notes
Observed logs suggest this can occur when multiple controllers/videos are present and one video is in lifecycle/initialization state, causing authoritative speed state to drift before cooldown fight-back stabilizes the active video.

MuizNadeem:
Fixes #1494

## Summary

This PR fixes a regression where keyboard relative speed shortcuts (`s` / `d`) could apply from a `1.0x` baseline after tab backgrounding, instead of from the active/effective remembered speed.

### Changes

- Use remembered non-`1.0` speed as baseline for the first relative step when:
  - current playback rate is effectively `1.0`, and
  - delta is a typical shortcut step.
- Treat lifecycle speed restoration as init-like (`source: 'init'`) in media event restore path.
- Do not persist `lastSpeed` for `source: 'init'` updates.

### Why

Lifecycle/init reconciliation should restore playback without mutating user-authoritative remembered speed. If lifecycle restore writes `lastSpeed`, first keyboard relative adjustment can compute from an incorrect `1.0` baseline after idle/background transitions.

## Test Plan

- [x] Added unit test: relative increase uses remembered baseline after idle reset to `1.0`.
- [x] Added unit test: play-event restoration does not overwrite `lastSpeed`.
- [ ] Manual:
  - Enable `rememberSpeed`
  - Set speed to `1.8x`
  - Background tab and return
  - Press `d`/`s`
  - Verify behavior is `1.9x` / `1.7x` (not `1.1x` / `0.9x`)

### DIFF
--- a/src/core/action-handler.js
+++ b/src/core/action-handler.js
@@ -408,17 +408,31 @@ class ActionHandler {
     // Calculate target speed
     let targetSpeed;
     if (relative) {
-      // For relative changes, add to current speed
+      // For relative changes, add to current speed. If the player has silently
+      // reset to 1.0 after idle, but we still have a non-1.0 in-memory speed,
+      // use that remembered baseline for the first relative step.
       const currentSpeed = video.playbackRate < 0.1 ? 0.0 : video.playbackRate;
-      targetSpeed = currentSpeed + value;
+      let baseSpeed = currentSpeed;
+      const rememberedSpeed = this.config?.settings?.lastSpeed;
+      const absDelta = Math.abs(value);
+      const isTypicalStepDelta = absDelta >= 0.05 && absDelta <= 0.5;
+      if (
+        isTypicalStepDelta &&
+        Math.abs(currentSpeed - 1.0) < 0.01 &&
+        typeof rememberedSpeed === 'number' &&
+        Math.abs(rememberedSpeed - 1.0) > 0.01
+      ) {
+        baseSpeed = rememberedSpeed;
+      }
+      targetSpeed = baseSpeed + value;
 
       // Snap to 1.0x when crossing the 1.0 boundary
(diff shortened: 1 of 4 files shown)


MuizNadeem:
Hi @igrigorik, could you please review this PR?

PR #1500, opened 2026-04-09
feat: add Crunchyroll support

galoryzen:
## Summary

Adds Crunchyroll support. Crunchyroll uses the Bitmovin player which causes two issues:

- **Overlay not visible**: The Bitmovin container has `overflow: hidden` and a `> *` CSS rule that collapses all direct children to 0x0. The controller is now inserted outside the container to escape clipping (same approach as the Netflix handler).
- **Speed resets on pause/resume and seek**: Bitmovin explicitly resets `playbackRate` on every resume — once synchronously and once via a Promise — and on seeks. This defeats the extension's fight-back mechanism. The handler saves the user's speed on `pause`/`seeking` events and restores it on `playing`/`seeked` events.

## Known limitations

- **Arrow key seek resets speed**: Crunchyroll's native seek (arrow keys) resets the speed to Bitmovin's default. This is because Bitmovin resets `playbackRate` synchronously before the browser's `seeking` event fires, so `savedSpeed` gets overwritten. The extension's own seek keys (Z/X) work correctly since they bypass Bitmovin entirely.
- **Skip intro / chapter transitions**: Seeking to certain timestamps (e.g. where the "Skip Intro" button appears) can cause Crunchyroll's React framework to re-render the player DOM, removing the controller and breaking the extension until page reload. This is a pre-existing issue with how the extension's mutation observer handles aggressive DOM re-renders, not specific to this handler.

## Test plan

- [ ] Load extension, open a Crunchyroll video page
- [ ] Confirm speed overlay appears on the video
- [ ] Set speed to 2x, pause, resume — speed should stay at 2x
- [ ] Change speed while paused, resume — new speed should persist
- [ ] Set Crunchyroll native speed to 0.75x, override with extension to 2x, pause, resume — should stay at 2x
- [ ] Seek with Z/X keys — speed should be preserved
- [ ] Confirm overlay is draggable and doesn't block Crunchyroll's player controls
- [ ] Confirm other sites (YouTube, Netflix) are unaffected

### DIFF
--- a/src/entries/inject-entry.js
+++ b/src/entries/inject-entry.js
@@ -44,6 +44,7 @@ import '../site-handlers/facebook-handler.js';
 import '../site-handlers/amazon-handler.js';
 import '../site-handlers/apple-handler.js';
 import '../site-handlers/dailymotion-handler.js';
+import '../site-handlers/crunchyroll-handler.js';
 import '../site-handlers/index.js';
 
 // Netflix-specific script
--- a/src/site-handlers/crunchyroll-handler.js
+++ b/src/site-handlers/crunchyroll-handler.js
@@ -1,10 +1,25 @@
 /**
  * Crunchyroll-specific handler
+ *
+ * Crunchyroll uses the Bitmovin player which:
+ * 1. Has overflow:hidden on its container, clipping the controller overlay
+ * 2. Explicitly resets playbackRate on every play/resume via its
+ *    setPlaybackSpeed chain — once synchronously and once via a Promise.
+ *    The synchronous reset overwrites lastSpeed before the extension's
+ *    play event handler runs. We save the speed on pause (before any
+ *    reset can happen) and restore on the playing event, which re-arms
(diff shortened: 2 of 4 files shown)


igrigorik:
Thanks for the contribution—the underlying Crunchyroll report is useful, but this implementation is no longer safe to merge. The branch predates the current speed-arbitration work: restoring from `playing`/`seeked` through `adjustSpeed()` mislabels lifecycle recovery as a user choice, can persist and advance shared authority, and duplicates core play/seek handling; the handler also shares one `savedSpeed` across videos and retains per-video listeners across player rerenders. Moving the controller outside Bitmovin’s container can remove it from the fullscreen subtree, and the PR adds no regression coverage. Current `master` now domain-scopes the old global `#player` rule that affected non-YouTube players and has substantially revised play, seek, and rate-change arbitration. Closing this version; please retest against current `master`, and if Crunchyroll still fails, open a fresh issue or PR with the current behavior so a narrow fix can be evaluated.

PR #1517, opened 2026-04-15
Codex/slider speed control

manuelruff:
Adds a small speed slider above the existing controller buttons and wires it into the current speed-change flow.

I found this useful for quicker direct speed selection, but I’m not sure whether this is something you’d want in the main controller UI, so I wanted to open it up for feedback first.

If this direction makes sense, this could also be made optional behind a setting so users can enable or disable it.

https://github.com/user-attachments/assets/6cf23687-fa91-498e-86df-9a994602fc2f

### DIFF
--- a/src/core/action-handler.js
+++ b/src/core/action-handler.js
@@ -505,6 +505,11 @@ class ActionHandler {
     }
     speedIndicator.textContent = numericSpeed.toFixed(2);
 
+    const speedSlider = video.vsc?.speedSlider;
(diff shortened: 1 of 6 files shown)


igrigorik:
Neat! In spirit of keep-it-simple, I'll lean on keyboard shortcuts as primary interface.

manuelruff:
Hi, Thanks for the responed! 
Ive made a fork with the changes i like :)

PR #1518, opened 2026-04-15
Add 1x quick reset button

manuelruff:
Adds a small `1x` button between the `-` and `+` controls so users can quickly return playback speed to normal.

I found this handy in practice, but I’m not sure whether it belongs in the default controller UI, so I wanted to put it up for feedback first.

If this direction makes sense, this could also be made optional behind a setting.

https://github.com/user-attachments/assets/d19aeb93-476f-4826-9af5-2f15827a819c

### DIFF
--- a/src/ui/controls.js
+++ b/src/ui/controls.js
@@ -17,7 +17,7 @@ class ControlsManager {
    */
   setupControlEvents(shadow, video) {
     this.setupDragHandler(shadow);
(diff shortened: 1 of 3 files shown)


igrigorik:
Yeah this is better handled as a custom shortcut: you can bind a custom key to one or more preferred speeds and use that to toggle.

PR #1532, opened 2026-04-29
Fix YouTube seek resetting native playback speed to 1x

### LINKED ISSUE #1521: Speed resets to default when unpausing or seeking
### Description

When the playback speed is adjusted using the website's **native speed control**, unpausing/moving the video progress bar causes the speed to reset to the default (1x), rather than retaining the user's selected speed.

### Steps to reproduce:

1. Use the website's native speed control to adjust the video playback speed (e.g., [YouTube](https://www.youtube.com/), set to 1.5x or 2x).
2. Move the video progress bar (seek to a different timestamp).
3. Observe the playback speed.

ColtonIdle:
Note: I verified the fix myself, but this change was done completely by AI, as well as the following description. This was mostly done to start a conversation with a possible solution. I'm an android dev so this chrome extension code goes over my head =)

## Summary

Fixes YouTube/native-player speed changes being lost after seeking or unpausing.

The extension previously treated any unhandled click/key as possible native speed-control intent. On YouTube, seeking can trigger a playbackRate reset to `1x` right after a progress-bar click, so the extension accepted that reset as the user's chosen speed.

This change narrows gesture tracking to likely native speed controls and known native speed shortcuts (`<` / `>`), so generic seek clicks continue to fight back to the selected speed.

Also allows a native speed-menu change to establish in-session `lastSpeed` even before the extension has set a speed itself.

Fixes #1521.

## Tests

- `npm test -- tests/unit/utils/event-manager.test.js`
- `npm run test:unit`
- `npm run lint -- src/utils/event-manager.js tests/unit/utils/event-manager.test.js`
- `npm run build`

### DIFF
--- a/src/utils/event-manager.js
+++ b/src/utils/event-manager.js
@@ -18,10 +18,9 @@ class EventManager {
     this.fightCount = 0;
     this.fightTimer = null;
 
-    // User gesture tracking: timestamp of the last user interaction we did NOT
-    // handle (click on page UI, unhandled key). A ratechange arriving within
-    // USER_GESTURE_WINDOW_MS of this is treated as intentional and accepted
-    // immediately rather than fought — handles native site speed controls.
+    // User gesture tracking: timestamp of the last likely native speed control
+    // interaction. A ratechange arriving within USER_GESTURE_WINDOW_MS of this
+    // is treated as intentional and accepted immediately rather than fought.
     this.lastUserInteractionAt = 0;
   }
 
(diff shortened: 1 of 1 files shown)


lucascehn:
I ran into this bug, realized this extension was the one causing the issue, tried the fix and it worked :-)

lucascehn:
Found another bug:
1. Play vid at 2x speed
2. Pause it
3. Go to a diff tab
4. Return back to vid
5. Unpause it
It'll play at 1x speed again

igrigorik:
Thanks for kicking off the conversation — the diagnosis was right. I went a
different route than DOM narrowing (YouTube's player DOM churns too often to
pin behavior to it): fixed in main generically — a reset to exactly 1.0x is
never adopted as user intent without strong evidence, so seek resets bounce
off everywhere, not just YouTube. Regression-tested. Closing in favor of that
approach.

PR #1563, opened 2026-06-07
Fix YouTube arrow seek speed reset

sharno:
Summary:
- Only treat unhandled keyboard gestures as native speed changes for speed shortcut keys like < and >.
- Keep YouTube arrow-key seeking from accepting a transient 1x reset as user intent.
- Add regression coverage for arrow seek reset and native speed shortcuts.

Tests:
- npm test -- tests/unit/utils/event-manager.test.js
- npm run test:unit
- npm test
- npx eslint "src/**/*.js" "tests/**/*.js"
- npm run build
- node tests/e2e/run-e2e.js youtube

### DIFF
--- a/src/utils/event-manager.js
+++ b/src/utils/event-manager.js
@@ -119,9 +119,12 @@ class EventManager {
         event.stopPropagation();
       }
     } else {
-      // Unhandled key — could be a site shortcut (e.g. YouTube's < > speed keys).
-      // Mark as user interaction so an immediately-following ratechange is accepted.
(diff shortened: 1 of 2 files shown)


igrigorik:
Thanks — this was exactly right, and main fixes it in your shape: only native
speed-shortcut keys (< / >) arm keyboard intent, so an arrow-key seek can't
get a transient 1x adopted. Your rule is cited in the classifier comments,
and arrow-seek resets have permanent regression coverage. Closing in favor of
the rework.

PR #1585, opened 2026-07-10
Fix settings bridge reinitialization and about:blank handling

YUMA-NAGAO:
## Summary

- keep the isolated-world settings bridge available across disable/re-enable cycles
- answer every settings handshake with the current sanitized settings snapshot
- evaluate disabled-site rules and per-site speed against the creator URL for inherited `about:` frames
- use that inherited context for domain CSS and site-handler selection as well
- avoid early-handshake and re-enable races, and block runtime relays/storage writes while inactive

## Testing

- `npm ci`
- `npm run lint`
- `npm run build:release`
- `npm test` (31 files, 396 tests)
- independent diff review and Codex second-opinion review

## Notes

- No manifest permissions changed.
- `npm audit --audit-level=high` reports the same existing 9 advisories (4 high) on `upstream/master` and this branch; dependency updates are intentionally out of scope.

### DIFF
--- a/src/content/inject.js
+++ b/src/content/inject.js
@@ -193,7 +193,7 @@ class VideoSpeedExtension {
    * non-matching rules never reach the browser's style invalidation engine.
    */
   preprocessDomainCSS(css) {
-    const hostname = location.hostname.replace(/^www\./, '');
+    const hostname = window.VSC.StorageManager.getContextHostname().replace(/^www\./, '');
     return css.replace(
       /:root\[style\*='--vsc-domain:\s*"([^"]+)"'\]([^{]*)\{([^}]*)\}/g,
       (match, domain, selector, body) => (domain === hostname ? `${selector.trim()} {${body}}` : '')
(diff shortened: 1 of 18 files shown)


igrigorik:
Thanks for surfacing the settings-handshake, disable/re-enable, and inherited-frame lifecycle problems. The core failures were reproduced, but this branch is now conflicting and its creator-URL plus site-handler expansion is broader than needed. A smaller maintainer-side implementation has been prepared that keeps re-enable reload-only, fails closed in inherited about: frames, blocks inactive relays and writes, and prevents deferred startup from surviving teardown, with unit and real-Chrome regression coverage. Closing this PR in favor of that narrower implementation.

YUMA-NAGAO:
@igrigorik Thanks for the update, and sorry I missed this earlier.

I was genuinely glad to hear that my report and investigation helped reproduce the core lifecycle issues and informed the narrower implementation.

If the project has a changelog, release notes, or another place where contributors are acknowledged, would you be open to briefly crediting @YUMA-NAGAO for initially surfacing and investigating these issues? It would mean a lot to me to have a visible record of the contribution.

I completely understand if this does not fit the project's usual process. Thanks again for taking the time to investigate and implement the fix.

Issue #1581, opened 2026-07-03
The video playback speed on Facebook is reset to 1.0x after dragging the time slider.


ngankvn:
When you click anywhere on the video progress bar, the playback speed will immediately change to 1.0.

Issue #1562, opened 2026-06-04
YouTube speed resets when using forward and backward keys.


SuperDazza:
When setting the YouTube speed specifically with YouTube's own speed controller (which I like to use because it also changes how far back and forward the video goes using the arrow keys) and then using the arrow keys to move forwards or backwards, Video Speed Controller resets the video's speed back to 1.00.

gabrielrdrguez:
happening to me aswell

igrigorik:
Fixed in main, ships in the next release. Only native speed-shortcut keys
(< / >) arm keyboard intent now, so an arrow-key seek can't get a transient 1x
adopted (thanks @sharno — #1563). Permanent regression coverage added.

gabrielrdrguez:
when can we expect then next release?

PR #1603, opened 2026-08-06
Stop click evidence from adopting a site reset to exactly 1.0

### LINKED ISSUE #1600: Keeping the selected speed (Remember playback speed) after interacting with the video is broken
The update to v0.11.0 basically broke the "Remember playback speed" setting (again).

Some videos do not even start in the chosen speed, some do.
Interacting with any video like skipping time resets the playback speed to 1.0 with every video.

This was an issue in the past, maybe the same bug is back or another got introduced.

Workaround: Pressing the G (preferred speed) key. You have to do that with every video after each time skip though.

Nahid-NHB:
Fixes #1600.

## What's broken

"Remember playback speed" dies as soon as you touch a video on 0.11.0. The reporter's verbose log has the whole story in two lines:

```
INFO:initializeSpeed: restoring speed to 2.1
INFO:Accepting site speed change as user-intentional: 1 (input 226ms ago)
```

The player's own reset to `1.0` came back classified `USER_INTENT`. That's the one verdict that emits `PERSIST`, so it overwrote `lastSpeed` globally. That's why every *later* video also started at 1.0, not just the one being watched, and why pressing G was the only workaround.

## Why the classifier said "user intent"

The STRONG tier could be reached by a click *sequence*: any two clicks within 5s, the last one inside the 300ms gesture window. STRONG skipped the exact-1.0 guard that the WEAK (single click) tier enforces.

The assumption in b583aeb was that real speed menus are >= 2 clicks deep, so a 2-click sequence means menu traversal. That part is true. The converse isn't: click play, click the progress bar, the player resets the rate on seek, and it reads as "user picked Normal." Same failure class as #1581 and #1562, just arriving through the sequence path instead of the single-click path.

## The change

No click tier makes a transition to exactly 1.0 durable anymore. `strong` splits into `speedKey` and `clickSequence`:

- `speedKey` (native `<` / `>`) names the speed control directly, so it still adopts any value, 1.0 included. It stays the one durable path to normal.
- `clickSequence` keeps its rank above the hold signature, so a native menu pick of 2x isn't swallowed by a concurrent press, but it now falls under the same `!isNormalReset` guard as a single click.

Clicks can't separate "menu -> Normal" from "play, then seek." The observations are identical, and no signal available to the classifier distinguishes them. So the tie breaks on cost instead of likelihood: adopting wrongly rewrites `lastSpeed` for every future video on every site, while fighting wrongly spends one media element's bounded fight budget and then surrenders locally with `lastSpeed` intact.

## Tradeoff

Choosing "Normal" in a site's own speed menu is now fought ~4 times before that video settles at 1.0, instead of adopting instantly. It's visible, but it's bounded, it self-resolves, and VSC's own reset key fixes it in one keystroke. Losing your remembered speed everywhere isn't recoverable that way.

## Arbitration contract

Per CONTRIBUTING: no transition-table cell changes. The arbiter, the contract table and `specs/SpeedArbiter.tla` are untouched. This lives entirely in the classifier, which the model already treats as an untrusted oracle. `docs/speed-arbitration.md` is updated for the new tier semantics.

## Verification

- New `#1600` bug-ledger entry in the differential suite: fails on master with `{rate: 1, mem: 1}`, passes here with `{rate: 2, mem: 2}`. Both the live pipeline and the pure model.
- `npm test` — 537 passed, 0 failed. Three existing assertions changed: all three asserted the old adopt-1.0-on-sequence behavior. The YouTube hold-release test now asserts `hasStrongIntent` is still true while the verdict is autonomous, so it still proves the release guard rather than the classifier verdict.
- `npm run test:tlc` — no error found, 49152 distinct states.
- Lint and prettier clean.
- Full E2E sweep against real Chrome, every suite green: basic 10/10, settings 4/4, display 6/6, lifecycle 2/2, arbitration 6/6, youtube 13/13. That includes both YouTube hold tests ("temporary 2x hold restores the VSC speed on release" and the missing-release-ratechange fallback), which are the ones this change could plausibly disturb.

One local-environment note in case anyone else hits it: launching the headful E2E browsers back to back, or overlapping two runs, made whole suites fail here with `Attempted to use detached Frame`, on master exactly as much as on this branch. Leaving a few seconds between runs and not overlapping them made it go away. Nothing to fix in the repo, just don't let two of these browsers run at once.


igrigorik:
Superseded by https://github.com/igrigorik/videospeed/commit/39f6b33ab8d1ae618dcd62d62ba2d7dea261410e.

